PR #82, opened 2025-06-27
[misc] fix helm chart and config

slin1237:


CatherineSue (dismissed):

charts/ome-resources/values.yaml
```diff
@@ -93,53 +72,20 @@ ome:
       memoryLimit: 320Gi
       cpuRequest: 15
       cpuLimit: 15
```
beiguo218, line 74: Do we need to remove all values here for fineTunedAdapter?

config/models/deepseek-ai/DeepSeek-V3.yaml
```diff
@@ -6,25 +6,6 @@ spec:
   vendor: deepseek-ai
   disabled: false
   version: "1.0.0"
-  modelFormat:
-    name: safetensors
```
pallasathena92, line 10: why remove these modelFormat, modelFramework, modelArchitecture in model cr?

PR #89, opened 2025-06-28
[BUG] Fix Inference Service Status Propagation

YouNeedCryDear:
## What type of PR is this?
Bug
Cleanup

1. Fix inference service status propagation. 
    a. Previous the status will be ready regardless of the other components. 
    b. Right now the status of inference service will wait for all the components to be ready then change to ready. 
2. Fix sgl runtimes, possibly due to previous rebase error
3. Remove unused FP8 models and runtimes

Fixes #

No

```release-note

```


YouNeedCryDear:
@gemini-code-assist review

pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
```diff
@@ -241,6 +241,51 @@ func (sr *StatusReconciler) SetModelFailureInfo(status *v1beta1.InferenceService
...
+			readyCondition.Reason = string(component) + "NotReady"
+			// Get the actual condition to propagate the message
+			compCond := status.GetCondition(componentReadyCondition)
+			if compCond != nil {
+				readyCondition.Message = compCond.Message
+			} else {
+				readyCondition.Message = "Component " + string(component) + " is not ready"
+			}
+			break // one not ready is enough
+		}
```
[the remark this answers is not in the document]
YouNeedCryDear: I think the purpose is to make sure the inference service status only change to ready when all the components are ready no matter it's one or many. Also the reconciler logic should pick up all the components which are not ready.

config/runtimes/srt/deepseek-rdma-pd-rt.yaml
```diff
@@ -63,7 +63,7 @@ spec:
             MC_TE_METRIC=true;
             SGLANG_TBO_DEBUG=1;
             python3 -m sglang.launch_server
-            --port 30000
```
slin1237: lets have those config changes in a different PR

pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
```diff
@@ -241,6 +241,51 @@ func (sr *StatusReconciler) SetModelFailureInfo(status *v1beta1.InferenceService
 }
 
 // PropagateCrossComponentStatus aggregates conditions across components
+// AggregateComponentReadyCondition creates the top-level Ready condition
+// based on the readiness of all specified components.
+func (sr *StatusReconciler) AggregateComponentReadyCondition(
```
slin1237: i think we already have something similar, since we use engine
we should enforce status to be ready if and only if ingress and engine are ready, check the api code

slin1237 (changes requested):

config/runtimes/srt/deepseek-rdma-pd-rt.yaml
HaiShaw: second.

pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
YouNeedCryDear: Shouldn't we enforce the status to be ready after every component is ready? For example PD, both the engine and decoder has to be ready before the inference service can be declared ready.

config/default/manager_image_patch.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     spec:
       containers:
         - name: manager
-          image: ghcr.io/moirai-internal/ome-manager:v0.1.0
+          image: ghcr.io/moirai-internal/ome-manager:v0.1.1
```
slin1237:
```suggestion
          image: ghcr.io/moirai-internal/ome-manager:v0.1.2
```

config/default/model_agent_image_patch.yaml
```diff
@@ -8,4 +8,4 @@ spec:
     spec:
       containers:
         - name: model-agent
-          image: ghcr.io/moirai-internal/model-agent:v0.1.0
+          image: ghcr.io/moirai-internal/model-agent:v0.1.1
```
slin1237:
```suggestion
          image: ghcr.io/moirai-internal/model-agent:v0.1.2
```

slin1237 (approved):

PR #129, opened 2025-07-04
[oep] add proposal for accelerator aware runtime selection

slin1237:
## What type of PR is this?

  /kind feature
  /kind design

  This PR propose OEP-0003: Accelerator-Aware Runtime Selection for Heterogeneous GPU Environments.

  ### Problem
  Currently, OME's runtime selection mechanism lacks awareness of underlying hardware accelerators. In clusters with mixed GPU types (H100, A100, B200, H200), operators
   must create separate runtime configurations for each GPU model, leading to:
  - Runtime proliferation (e.g., sglang-h100, sglang-a100, sglang-b200, sglang-h200)
  - Operational complexity and configuration drift
  - Suboptimal GPU placement and resource utilization
  - Difficulty managing PD-disaggregated deployments with different GPU types

  ### Solution
  This PR introduces AcceleratorClass, a vendor-agnostic abstraction layer that:
  - Defines GPU capabilities and requirements declaratively
  - Enables automatic runtime selection based on model needs and GPU availability
  - Supports cost optimization and performance policies
  - Integrates seamlessly with existing Kueue ResourceFlavors
  - Works with OME's engine/decoder/router architecture

  ### Key Components
  1. **AcceleratorClass CRD**: Cluster-scoped resource defining GPU types and capabilities
  2. **Runtime Extensions**: AcceleratorRequirements field for GPU compatibility
  3. **InferenceService Extensions**: AcceleratorSelector for GPU selection preferences
  4. **Smart Selection Logic**: Matches models to optimal GPUs based on memory, compute, and features
  5. **Kueue Integration**: Optional synchronization with ResourceFlavors

  ### Example Usage
  ```yaml
  # Define GPU type
  apiVersion: ome.io/v1beta1
  kind: AcceleratorClass
  metadata:
    name: nvidia-h100
  spec:
    deviceSelector:
      deviceResourceName: "nvidia.com/gpu"
      nodeLabelSelector:
        nvidia.com/gpu.product: "NVIDIA-H100-80GB-HBM3"
    capabilities:
      memoryCapacity: 80
      computeCapability: 1979

  ---
  # Deploy with automatic GPU selection
  apiVersion: ome.io/v1beta1
  kind: InferenceService
  metadata:
    name: llama-70b
  spec:
    model:
      name: llama-70b
    acceleratorSelector:
      policy: BestFit  # Automatically selects appropriate GPU
```

###  Architecture Decisions

  1. Cluster-scoped AcceleratorClass: GPUs are cluster resources, making cluster scope appropriate
  3. Kueue Optional: Integration is optional to avoid forcing Kueue adoption
  4. Backward Compatible: All new fields are optional, existing InferenceServices work unchanged
  5. Component-level Overrides: Engine and Decoder can have different accelerators for PD-disaggregated deployments
  6. Router Exclusion: Router components remain CPU-only as they don't require GPU resources

  

  Added AcceleratorClass resource for GPU-aware runtime selection in heterogeneous environments.

  - New AcceleratorClass CRD to define GPU types and capabilities
  - Extended InferenceService with acceleratorSelector for GPU selection
  - Added acceleratorRequirements to ServingRuntime for GPU compatibility
  - Automatic GPU discovery and runtime matching
  - Support for constraint-based and policy-driven GPU selection
  - Optional Kueue integration for ResourceFlavor synchronization
  - Component-level accelerator overrides for PD-disaggregated deployments
  - Comprehensive examples and migration guide

### DIFF
--- a/oeps/0003-accelerator-aware-runtime-selection/README.md
+++ b/oeps/0003-accelerator-aware-runtime-selection/README.md
@@ -1,0 +1,1655 @@
+# OEP-0003: Accelerator-Aware Runtime Selection for Heterogeneous GPU Environments
+
+<!--
+This OEP proposes a new accelerator abstraction layer for OME to better handle
+heterogeneous GPU environments and reduce runtime proliferation while maintaining
+compatibility with existing Kubernetes ecosystem tools like Kueue.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: ML Practitioner Deploying Models](#story-1-ml-practitioner-deploying-models)
+    - [Story 2: Platform Engineer Managing Multiple GPU Types](#story-2-platform-engineer-managing-multiple-gpu-types)
+    - [Story 3: Heavy Kueue User](#story-3-heavy-kueue-user)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Specifications](#api-specifications)
+    - [AcceleratorClass](#acceleratorclass)
+    - [InferenceService Extensions](#inferenceservice-extensions)
+    - [ServingRuntime Extensions](#servingruntime-extensions)
+  - [Kueue Integration](#kueue-integration)
+  - [Override Hierarchy](#override-hierarchy)
+  - [Implementation Architecture](#implementation-architecture)
+  - [Test Plan](#test-plan)
+    - [Unit Tests](#unit-tests)
+    - [Integration Tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This OEP introduces an AcceleratorClass abstraction to OME that enables intelligent runtime selection in heterogeneous GPU environments. Currently, supporting multiple GPU types (e.g., H100, A100, B200, H200) requires creating numerous runtime configurations - a combinatorial explosion that becomes unmanageable. The proposed solution provides a vendor-agnostic way to define accelerator capabilities and automatically match them with appropriate runtimes.
+
+The design integrates seamlessly with existing Kubernetes ecosystem tools, particularly Kueue's ResourceFlavor concept, enabling users to leverage existing resource management infrastructure. By introducing capability-based matching rather than hard-coding specific GPU models, the system remains flexible and future-proof as new accelerator types emerge.
+
+## Motivation
+
+OME's current runtime selection mechanism matches runtimes based on model characteristics (format, architecture, size, quantization) but lacks awareness of the underlying hardware accelerators. In clusters with heterogeneous GPU types, this limitation forces operators to create and maintain separate runtime definitions for each GPU model - leading to operational complexity and configuration drift.
+
+Furthermore, the lack of standardization across GPU vendors (NVIDIA, AMD, Intel) in naming conventions and resource exposure makes it challenging to build a unified solution. Each vendor uses different labeling schemes, and Kubernetes device plugins expose resources differently (e.g., nvidia.com/gpu vs amd.com/gpu).
+
+### Goals
+
+1. **Reduce runtime proliferation** - Enable a single runtime definition to work across multiple GPU types
+2. **Vendor-agnostic design** - Support NVIDIA, AMD, Intel, and future accelerators without code changes
+3. **Kueue integration** - Seamlessly work with existing Kueue ResourceFlavor deployments
+4. **Progressive disclosure** - Simple for basic use cases, powerful for advanced scenarios
+5. **Automatic optimization** - Select optimal GPU based on model requirements and availability
+6. **Clear override hierarchy** - Provide predictable configuration precedence
+
+### Non-Goals
+
+1. **GPU virtualization** - This OEP does not address GPU sharing or MIG configuration
+2. **Cost optimization** - While the design enables cost-aware scheduling, implementing cost models is out of scope
+3. **Dynamic runtime generation** - Automatically creating new runtime configurations based on discovered GPUs
+4. **Replacing existing APIs** - The design extends rather than replaces current runtime selection
+
+## Proposal
+
+Introduce new API resources and extensions to enable accelerator-aware runtime selection:
+
+1. **AcceleratorClass** (Cluster-scoped) - Defines accelerator capabilities and discovery patterns
+2. **Runtime/InferenceService extensions** - Add accelerator requirements and selection fields
+
+Note: AcceleratorProfile was considered but deferred to reduce initial complexity. The same functionality can be achieved through InferenceService's acceleratorSelector field.
+
+The system will automatically discover available accelerators, match them with runtime requirements, and select the optimal configuration based on model characteristics and user preferences.
+
+### User Stories
+
+#### Story 1: ML Practitioner Deploying Models
+
+Alice wants to deploy a Llama 7B model for inference. She doesn't know the differences between GPU types and just wants her model to run efficiently.
+
+**Current Experience:**
+```yaml
+# Alice deploys with auto-selected runtime
+kind: InferenceService
+metadata:
+  name: llama-7b
+spec:
+  model:
+    name: llama-7b
+  # Runtime is auto-selected based on model, but GPU type is not considered
+  # Result: Might get scheduled on any available GPU (A100, H100, etc.)
+  # Problem: No control over GPU selection, potentially inefficient resource use
+```
+
+**New Experience:**
+```yaml
+# Alice specifies only what she needs
+kind: InferenceService
+metadata:
+  name: llama-7b
+spec:
+  model:
+    name: llama-7b
+  runtime:
+    name: sglang-universal  # Automatically selects appropriate GPU
+```
+
+#### Story 2: Platform Engineer Managing Multiple GPU Types
+
+Bob manages a cluster with A100-40GB, A100-80GB, H100-80GB, and H200-96GB GPUs. He needs to support multiple model architectures without creating 16+ runtime configurations.
+
+**Current Experience:**
+```yaml
+# Bob creates a runtime that works on all GPUs but can't optimize per GPU type
+kind: ServingRuntime
+metadata:
+  name: sglang-all-gpus
+spec:
+  supportedModelFormats:
+    - name: safetensors
+  containers:
+  - name: sglang
+    env:
+    # Must use conservative settings that work on smallest GPU
+    - name: GPU_MEMORY_UTILIZATION
+      value: "0.85"  # Conservative for all GPUs
+    - name: MAX_MODEL_LEN  
+      value: "16384" # Limited by smallest GPU
+    # Can't enable H100-specific optimizations like FP8
+```
+
+**New Experience:**
+```yaml
+# Bob creates one universal runtime
+kind: ServingRuntime
+metadata:
+  name: sglang-universal
+spec:
+  supportedModelFormats:
+    - name: safetensors
+      autoSelect: true
+  
+  acceleratorRequirements:
+    requiredCapabilities:
+      minComputeCapability: "8.0"
+      requiredFeatures: ["tensor-cores"]
+    
+  # GPU-specific optimizations
+  acceleratorConfigurations:
+    - selector:
+        acceleratorClass: nvidia-a100-40gb
+      env:
+        - name: TENSOR_PARALLEL_SIZE
+          value: "2"
+        - name: GPU_MEMORY_UTILIZATION
+          value: "0.90"
+        - name: MAX_MODEL_LEN
+          value: "16384"
+      runner:
+        args:
+          - "--enable-prefix-caching"
+          
+    - selector:
+        acceleratorClass: nvidia-h100-80gb
+      env:
+        - name: TENSOR_PARALLEL_SIZE
+          value: "1"
+        - name: ENABLE_FP8
+          value: "true"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: oeps/0003-accelerator-aware-runtime-selection/README.md [hunk 1]
```diff
@@ -0,0 +1,1655 @@
...
+      minComputeCapability: "8.0"
+      requiredFeatures: ["tensor-cores"]
+    
+  # GPU-specific optimizations
+  acceleratorConfigurations:
+    - selector:
+        acceleratorClass: nvidia-a100-40gb
+      env:
+        - name: TENSOR_PARALLEL_SIZE
+          value: "2"
```
YouNeedCryDear, line 156: If this sglang-universal runtime is model agnostic, how does it know what is the expected tensor parallel size before model is specified?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+        - name: GPU_MEMORY_UTILIZATION
+          value: "0.90"
+        - name: MAX_MODEL_LEN
+          value: "16384"
+      runner:
+        args:
+          - "--enable-prefix-caching"
+          
+    - selector:
+        acceleratorClass: nvidia-h100-80gb
```
YouNeedCryDear, line 166: Is it possible to have multiple acceleratorClass with the same name, say nvidia-h100-80gb, but different env and runner specs?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+    nvidia.com/gpu.product: "A100-SXM4-40GB"
+
+---
+# OME automatically discovers and creates AcceleratorClass
+apiVersion: ome.io/v1beta1
+kind: AcceleratorClass
+metadata:
+  name: gpu-nvidia-a100
+  annotations:
+    ome.ai/kueue-resource-flavor: "gpu-nvidia-a100"
```
YouNeedCryDear: ome.io?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+  name: cost-sensitive-deployment
+spec:
+  model:
+    name: llama-13b
+  runtime:
+    name: sglang-cost-optimized
+  acceleratorSelector:
+    strategy: cost  # Prefer cost over performance
+    annotations:
+      ome.ai/cost-threshold: "0.5"  # $/hour threshold
```
YouNeedCryDear: assuming this cost-threshold will be defined on node level? Then it will be divided to per accelerator/GPU  level?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+      value: "16384" # Limited by smallest GPU
+    # Can't enable H100-specific optimizations like FP8
+```
+
+**New Experience:**
+```yaml
+# Bob creates one universal runtime
+kind: ServingRuntime
+metadata:
+  name: sglang-universal
```
YouNeedCryDear, line 139: For a single runtime like sglang-universal, those different acceleratorRequirements underneath might not necessarily produce the same level of performance for a given model. Would it be a problem if model randomly got allocated to different accelerator settings but the performance is not consistent?

YouNeedCryDear (commented):
From a high level perspective, I agree that the explosive combination of model and GPU type is not manageable. I wonder instead of introducing the accelerator suite, which seems like a pretty big change to both the cluster management as well as orchestration, would it be possible to have one runtime to represent one type of GPU, and leave the heavy lifting, such as the resource management to inference service? Basically just get rid of the model from current runtime.

oeps/0003-accelerator-aware-runtime-selection/README.md
slin1237: its still one runtime per model

oeps/0003-accelerator-aware-runtime-selection/README.md
slin1237: no

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+          --torch-compile-max-bs 1 
+          --reasoning-parser deepseek-r1 
+          --enable-metrics
+  # Even with acceleratorSelector, args from acceleratorConfigurations are NOT applied
+  # because user specified command
+  acceleratorSelector:
+    preferredClasses: ["nvidia-h100-80gb"]
+```
+
+**Scenario 2: User provides args array (accelerator args are appended)**
```
pallasathena92, line 479: we will not appended accelerator args only when customer provide command by themself?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+  family: ponte-vecchio
+  model: max-1550
+  
+  discovery:
+    nodeSelector:
+      gpu.intel.com/product: "max-1550"
+  
+  capabilities:
+    memoryGB: "128Gi"
+    levelZeroVersion: "1.3"
```
YouNeedCryDear, line 743: What is this? I don't see this attribute defined in the AcceleratorCapabilities struct.

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+    // PCIVendorID for device discovery (e.g., "10de" for NVIDIA)
+    // +optional
+    PCIVendorID string `json:"pciVendorID,omitempty"`
+    
+    // DeviceIDs list of PCI device IDs
+    // +optional
+    DeviceIDs []string `json:"deviceIDs,omitempty"`
+}
+
+type AcceleratorCapabilities struct {
```
YouNeedCryDear, line 968: Do we want this to be a pre-defined struct or a free form list?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+    // +optional
+    MemoryBandwidthGBps *resource.Quantity `json:"memoryBandwidthGBps,omitempty"`
+    
+    // Features supported by this accelerator
+    // +optional
+    Features []string `json:"features,omitempty"`
+    
+    // Performance metrics
+    // +optional
+    Performance *AcceleratorPerformance `json:"performance,omitempty"`
```
YouNeedCryDear, line 991: I think AcceleratorPerformance is missing definition.

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+    Class string `json:"class,omitempty"`
+    
+    // Score for this preference (higher is better)
+    // +kubebuilder:validation:Minimum=0
+    // +kubebuilder:validation:Maximum=100
+    Score int32 `json:"score"`
+    
+    // Conditions when this preference applies
+    // +optional
+    Conditions []PreferenceCondition `json:"conditions,omitempty"`
```
YouNeedCryDear, line 1157: Missing definition for PreferenceCondition?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+spec:
+  supportedModelFormats:
+    - name: safetensors
+      autoSelect: true
+  
+  # Define cost-aware preferences
+  acceleratorRequirements:
+    preferenceOrder:
+      - class: nvidia-a100-40gb
+        score: 100
```
YouNeedCryDear, line 322: How the score and conditions are applied? condition takes first and then pick the higher score?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+  name: cost-sensitive-deployment
+spec:
+  model:
+    name: llama-13b
+  runtime:
+    name: sglang-cost-optimized
+  acceleratorSelector:
+    strategy: cost  # Prefer cost over performance
+    annotations:
+      ome.io/cost-threshold: "0.5"  # $/hour threshold
```
YouNeedCryDear, line 364: Is this the total cost for the inference service? I still don't have a clear idea how this is calculated or applied.

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+1. **GPU virtualization** - This OEP does not address GPU sharing or MIG configuration
+2. **Cost optimization** - While the design enables cost-aware scheduling, implementing cost models is out of scope
+3. **Dynamic runtime generation** - Automatically creating new runtime configurations based on discovered GPUs
+4. **Replacing existing APIs** - The design extends rather than replaces current runtime selection
+
+## Proposal
+
+Introduce new API resources and extensions to enable accelerator-aware runtime selection:
+
+1. **AcceleratorClass** (Cluster-scoped) - Defines accelerator capabilities and discovery patterns
```
YouNeedCryDear, line 70: We probably need some RBAC or mutation webhook to control this AcceleratorClass doesn't get modified by unexpected parties.

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+    RouterConfig         *RouterSpec             `json:"routerConfig,omitempty"`
+    EngineConfig         *EngineSpec             `json:"engineConfig,omitempty"`
+    DecoderConfig        *DecoderSpec            `json:"decoderConfig,omitempty"`
+    
+    // NEW: Accelerator requirements for this runtime
+    // +optional
+    AcceleratorRequirements *AcceleratorRequirements `json:"acceleratorRequirements,omitempty"`
+}
+
+type AcceleratorRequirements struct {
```
YouNeedCryDear, line 1132: Normally LLM tensor parallelism depends on aggregated memory across multiple shards. Would it be a good idea to add something like minTotalMemoryGB?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+    
+    // NodeSelector for specific node targeting
+    // +optional
+    NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+    
+    // Strategy for selection (performance, cost, balanced)
+    // +kubebuilder:default="balanced"
+    // +optional
+    Strategy AcceleratorSelectionStrategy `json:"strategy,omitempty"`
+}
```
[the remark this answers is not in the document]
pallasathena92: there needs update

oeps/0003-accelerator-aware-runtime-selection/README.md
EdHasNoLife: Out of curiosity, what would happen if the cost passes the threshold?

oeps/0003-accelerator-aware-runtime-selection/README.md
slin1237: this is intelGPU lol

oeps/0003-accelerator-aware-runtime-selection/README.md
slin1237: How scoring and conditions work:
    - Conditions are applied FIRST as filters - they eliminate incompatible options
    - Scoring is applied SECOND to rank remaining options - higher scores are preferred

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+
+Furthermore, the lack of standardization across GPU vendors (NVIDIA, AMD, Intel) in naming conventions and resource exposure makes it challenging to build a unified solution. Each vendor uses different labeling schemes, and Kubernetes device plugins expose resources differently (e.g., nvidia.com/gpu vs amd.com/gpu).
+
+### Goals
+
+1. **Reduce runtime proliferation** - Enable a single runtime definition to work across multiple GPU types
+2. **Vendor-agnostic design** - Support NVIDIA, AMD, Intel, and future accelerators without code changes
+3. **Kueue integration** - Seamlessly work with existing Kueue ResourceFlavor deployments
+4. **Progressive disclosure** - Simple for basic use cases, powerful for advanced scenarios
+5. **Automatic optimization** - Select optimal GPU based on model requirements and availability
```
frankzhouhr, line 56: How is "optimal" defined? I think there's an opportunity for GPU fragmentation reduction here. Things like prefer partially used GPUs in placement algorithm

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+
+### Kueue Integration
+
+The system provides bi-directional integration with Kueue ResourceFlavors:
+
+```go
+// AcceleratorClassController watches for Kueue ResourceFlavors
+func (r *AcceleratorClassReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+    // Check if this is a Kueue ResourceFlavor
+    if req.Namespace == "" && strings.HasPrefix(req.Name, "gpu-") {
```
beiguo218, line 1169: nit: use hardcoded prefix to identify specific resource flavor may not be that flexible, probably we can switch to use label for it?

oeps/0003-accelerator-aware-runtime-selection/README.md
```diff
@@ -0,0 +1,1655 @@
...
+    // For engine and decoder components, apply accelerator constraints
+    
+    // 1. Start with AcceleratorClass nodeSelector
+    if acceleratorClass != nil && acceleratorClass.Spec.Discovery.NodeSelector != nil {
+        for k, v := range acceleratorClass.Spec.Discovery.NodeSelector {
+            merged[k] = v
+        }
+    }
+    
+    // 2. Merge Runtime nodeSelector (runtime can override accelerator)
```
beiguo218, line 1349: Should we let the accelerator's node selector override the runtime default one?

PR #139, opened 2025-07-06
fix: remove GPU node selector from benchmark jobs

abatilo:
Benchmark jobs execute performance tests against inference services and
don't require GPU resources themselves. Removing the nvidia.com/gpu node
selector allows these jobs to run on any available node, improving
scheduling flexibility and resource utilization.


slin1237:
GenAI-bench requires tokenizer
That tokenizer is currently only available in GPU nodes
the problem is probably how we handled getting the tokenizer benchmark controller

PR #148, opened 2025-07-08
add modelFormat, modelFramework version compare

pallasathena92:
## What type of PR is this?

/kind feature

Summary
This PR enhances the runtime compatibility check system by implementing a more scalable approach to comparing model formats and versions. The previous implementation used basic string comparison for model formats, frameworks, and versions, which was insufficient for proper semantic version handling and future extensibility.

Changes
Added proper semantic version parsing and comparison using semver principles
Implemented support for version comparison operators
Added special handling for unofficial versions (pre-release, build metadata)
Created dedicated unit tests in a separate file to isolate and verify the version comparison logic

Benefits
More reliable version comparison that follows semantic versioning principles
Support for comparing versions with different operators (equal to, greater than)
Special handling for pre-release and build metadata versions
Improved compatibility determination between models and runtimes
Better extensibility for future model format support

The changes are internal to the controller logic and affect how the system determines compatibility between models and runtimes. While users might observe more accurate matching between models and runtimes (particularly when semantic versioning is involved), they don't need to change how they interact with the system.

```release-note

```


pkg/controller/v1beta1/inferenceservice/utils/runtime.go
```diff
@@ -267,6 +255,140 @@ func RuntimeSupportsModel(baseModel *v1beta1.BaseModelSpec, srSpec *v1beta1.Serv
...
+
+			// Check if versions have unofficial parts (requirement #1)
+			hasUnofficialFormatVersion = containsUnofficialVersion(baseModelFormatVersion) ||
+				containsUnofficialVersion(supportedFormatVersion)
+
+			// Compare versions based on operator and whether unofficial versions exist (requirements #1, #2, #3)
+			if hasUnofficialFormatVersion || operator == "Equal" {
+				modelFormatMatches = Equal(supportedFormatVersion, baseModelFormatVersion)
+			} else if operator == "GreaterThan" {
+				modelFormatMatches = GreaterThan(supportedFormatVersion, baseModelFormatVersion)
```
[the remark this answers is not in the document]
pallasathena92: we want runtime version greater than model's version
in function
```
func GreaterThan(v Version, o Version) bool {
	return CompareVersion(v, o) == 1
}
```
it check if v large than o. In order to make logic easily to check,  put supportedFormatVersion as the first argument

pkg/apis/ome/v1beta1/servingruntime_types.go
```diff
@@ -49,8 +49,20 @@ type SupportedModelFormat struct {
 	// Priority can be overridden by specifying the runtime in the InferenceService.
 	// +optional
 	Priority *int32 `json:"priority,omitempty"`
+
+	// Operator for the selector with supported values: "Equal", "GreaterThan"
+	// This is used to select the serving runtime based on the modelFormat version, modelFramework version
+	// +optional
+	Operator *RuntimeSelectorOperator `json:"operator,omitempty"`
```
slin1237: this should be in both framework and format version? this should not be here

pkg/apis/ome/v1beta1/servingruntime_types.go
```diff
@@ -49,8 +49,20 @@ type SupportedModelFormat struct {
 	// Priority can be overridden by specifying the runtime in the InferenceService.
 	// +optional
 	Priority *int32 `json:"priority,omitempty"`
+
+	// Operator for the selector with supported values: "Equal", "GreaterThan"
```
slin1237: add a default value `// +kubebuilder:default=Equal`

pkg/apis/ome/v1beta1/servingruntime_types.go
```diff
@@ -49,8 +49,20 @@ type SupportedModelFormat struct {
...
+	// This is used to select the serving runtime based on the modelFormat version, modelFramework version
+	// +optional
+	Operator *RuntimeSelectorOperator `json:"operator,omitempty"`
 }
 
+type RuntimeSelectorOperator string
+
+const (
+	RuntimeSelectorOpEqual       RuntimeSelectorOperator = "Equal"
```
slin1237: add `GreaterOrEqual`
so
```go
type RuntimeSelectorOperator string

const (
	RuntimeSelectorOpEqual          RuntimeSelectorOperator = "Equal"
	RuntimeSelectorOpGreaterThan    RuntimeSelectorOperator = "GreaterThan"
	RuntimeSelectorOpGreaterOrEqual RuntimeSelectorOperator = "GreaterOrEqual"
)
```

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+type Version struct {
+	Major uint64
+	Minor uint64
+	Patch uint64
+	Pre   []string
+	Build []string //No Precedence
+	Dev   []string
+}
+
+// we will only support these three type of version
```
slin1237:
```
// Parse parses version strings like:
//   - 4.51.3-SAM-HQ-preview
//   - 4.43.0.dev0
//   - 4.43.0+build
//   - 0.6.0
```

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+	parts := strings.SplitN(s, ".", 3)
+	if len(parts) != 3 {
+		return Version{}, errors.New("No Major.Minor.Patch elements found")
+	}
+
+	// Major
+	if !containsOnly(parts[0], numbers) {
+		return Version{}, fmt.Errorf("Invalid character(s) found in major number %q", parts[0])
+	}
+	if hasLeadingZeroes(parts[0]) {
```
slin1237: both major and minor need this functionality
perhaps
```go
func parseNumeric(s, field string) (uint64, error) {
	if !containsOnly(s, numbers) {
		return 0, fmt.Errorf("invalid character(s) in %s number %q", field, s)
	}
	if hasLeadingZeroes(s) {
		return 0, fmt.Errorf("%s must not have leading zeroes: %q", field, s)
	}
	return strconv.ParseUint(s, 10, 64)
}
```

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	if hasLeadingZeroes(parts[1]) {
+		return Version{}, fmt.Errorf("Minor number must not contain leading zeroes %q", parts[1])
+	}
+	minor, err := strconv.ParseUint(parts[1], 10, 64)
+	if err != nil {
+		return Version{}, err
+	}
+
+	v := Version{}
```
slin1237: nit
```go
v := Version{Major: major, Minor: minor}
```

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+	v.Major = major
+	v.Minor = minor
+
+	var build, prerelease, dev []string
```
slin1237: nit
```
var (
    build, pre, dev []string
)
```

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+	patchStr := parts[2]
+
+	if buildIndex := strings.IndexRune(patchStr, '+'); buildIndex != -1 {
```
slin1237: nit
leave comment `// Extract +build`

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+		prerelease = strings.Split(patchStr[preIndex+1:], ".")
+		patchStr = patchStr[:preIndex]
+	}
+
+	if devIndex := strings.IndexRune(patchStr, '.'); devIndex != -1 {
+		dev = strings.Split(patchStr[devIndex+1:], ".")
+		patchStr = patchStr[:devIndex]
+	}
+
+	if !containsOnly(patchStr, numbers) {
```
slin1237: same here, u can use `parseNumeric` function i mentioned above

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+
+func GreaterThan(v Version, o Version) bool {
+	return CompareVersion(v, o) == 1
+}
+
+// Compare compares Versions v to o:
+// -1 == v is less than o
+// 0 == v is equal to o
+// 1 == v is greater than o
+func CompareVersion(v Version, o Version) int {
```
slin1237:
```go
// CompareVersion returns -1 if v < o, 0 if equal, 1 if v > o
func CompareVersion(v, o Version) int {
	if cmp := compareUint64(v.Major, o.Major); cmp != 0 {
		return cmp
	}
	if cmp := compareUint64(v.Minor, o.Minor); cmp != 0 {
		return cmp
	}
	if cmp := compareUint64(v.Patch, o.Patch); cmp != 0 {
		return cmp
	}

	if cmp := compareStringSlices(v.Pre, o.Pre); cmp != 0 {
		return cmp
	}
	if cmp := compareStringSlices(v.Build, o.Build); cmp != 0 {
		return cmp
	}
	if cmp := compareStringSlices(v.Dev, o.Dev); cmp != 0 {
		return cmp
	}

	return 0
}
```
and helper function 
```go
func compareStringSlices(a, b []string) int {
	n := min(len(a), len(b))
	for i := 0; i < n; i++ {
		if a[i] != b[i] {
			return strings.Compare(a[i], b[i])
		}
	}
	return compareUint64(uint64(len(a)), uint64(len(b)))
}
```
and
```go
func compareUint64(a, b uint64) int {
	switch {
	case a > b:
		return 1
	case a < b:
		return -1
	default:
		return 0
	}
}
```

pkg/apis/ome/v1beta1/servingruntime_types.go
pallasathena92: Do we want to add this complexity here? I try to use this operator to control both framework and format. What do you think?

pkg/controller/v1beta1/inferenceservice/utils/runtime_utils.go
```diff
@@ -0,0 +1,234 @@
...
+	if len(v.Build) > 0 && len(o.Build) == 0 {
+		return -1
+	}
+
+	// compare all build metadata
+	for i := range v.Build {
+		if v.Build[i] != o.Build[i] {
+			return strings.Compare(v.Build[i], o.Build[i])
+		}
+	}
```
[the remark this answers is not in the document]
pallasathena92: in our case, version with build, pre-release, dev are special version, which needs to keep exactly equal.

pkg/apis/ome/v1beta1/servingruntime_types.go
slin1237: yes, it must be at every level, if we put them here, do we imply this to framework version, format version, or actual model version. what about fine tuning version. It's super confusing to users

slin1237 (approved):

PR #156, opened 2025-07-10
[Bugfix] try to get baseModelName and protocolVersion from Model and …

leprode:
…Engine when isvc.predictor is nil

## What type of PR is this?
/kind bug

In the inferenceService controller, the `Predictor` has been migrated to `Engine` and `Model`. Therefore, when I create a BenchmarkJob, it prints an error log like this:

```
InferenceService {namespace}/{isvc name} has no Model defined in Predictor spec
```

Fixes #

```release-note

```

### DIFF
--- a/pkg/controller/v1beta1/benchmark/controller.go
+++ b/pkg/controller/v1beta1/benchmark/controller.go
@@ -247,10 +247,14 @@ func (r *BenchmarkJobReconciler) createPodSpec(benchmarkJob *v1beta1.BenchmarkJo
 			return nil, err
 		}
 		var baseModelName string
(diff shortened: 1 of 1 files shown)


slin1237:
thanks for the fix. would you mind fix the CI too
thanks 🙏

leprode:
> thanks for the fix. would you mind fix the CI too thanks 🙏

I've fixed the panic bug in TestBenchmarkJobReconciler_reconcilePodSpec/successful_pod_spec_creation locally and applied the code assistant's suggestions.

It looks like I don't have permission to trigger CI. Please help trigger it and review the code again. Thanks!

PR #193, opened 2025-07-16
fix: don't panic when webhook fails to get deploy config

carlory:
## What type of PR is this?

/kind bug

See pr title.

Fixes #

```release-note
fix: don't panic when webhook fails to get deploy config
```

### DIFF
--- a/pkg/webhook/admission/isvc/inference_service_validation.go
+++ b/pkg/webhook/admission/isvc/inference_service_validation.go
@@ -28,7 +28,7 @@ const (
 
 var (
 	// logger for the validation webhook.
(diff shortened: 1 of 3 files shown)


pkg/webhook/admission/isvc/inference_service_validation.go
```diff
@@ -28,7 +28,7 @@ const (
 
 var (
 	// logger for the validation webhook.
-	inferenceServiceValidatorLogger = logf.Log.WithName("inferenceservice-v1beta1-validation-webhook")
+	validatorLogger = logf.Log.WithName("inferenceservice-v1beta1-validation-webhook")
```
carlory, line 31: just rename for short.

slin1237 (approved):

PR #282, opened 2025-09-24
[Misc] Migrate to use XET-Core Rust based HF Hub in ome-agent replica

beiguo218:
## What type of PR is this?
/kind feature

This PR migrates the ome-agent replica from using the Go-based hfutil package to the new Rust-based XET-core library for Hugging Face Hub downloads. It preserves the existing interface while leveraging the performance benefits of the Rust implementation. Including changes for below:

1. Replaced `hfutil` with `xet` package in ome agent replica: Updated all Hugging Face download operations to use the new Rust-based XET-core client;
2. Modified the replica agent's dependency injection to use xet.Module instead of the old hub client;
3. Added/updated the unit tests.

Tested. Triggered a OME agent replica Job with 30 pods in parallel to import llama3.1 8b model from HF, no any 429 error being observed, all pods run successfully at their first try.


pkg/xet/libxet.a
slin1237, line 1: should we delete this or add this to gitignore

beiguo218:
Will do fix the build.

pkg/xet/libxet.a
beiguo218: Yea, would do.

beiguo218: Deleted

PR #317, opened 2025-11-20
Validate ClusterServingRuntime and ServingRuntime resources only reference known accelerator classes

Kangyan-Zhou:

Fixes #

```release-note

```


pkg/webhook/admission/servingruntime/servingruntime_webhook.go
```diff
@@ -295,3 +313,43 @@ func contains[T comparable](slice []T, element T) bool {
...
+	// Collect all missing classes to report them together
+	var missingClasses []string
+	for _, className := range spec.AcceleratorRequirements.AcceleratorClasses {
+		if !existingClasses[className] {
+			missingClasses = append(missingClasses, className)
+		}
+	}
+
+	if len(missingClasses) > 0 {
+		if len(missingClasses) == 1 {
```
pallasathena92: why we check length of missingClasses here? is there any special purpose?

pallasathena92: would we handle this error message without checking length of arrary？

Kangyan-Zhou: Yes simplify the code here, PTAL

PR #329, opened 2025-11-25
[Core] Update to only append an env variable when it is not provided in runtime/runner

beiguo218:
## What type of PR is this?
/kind Improvement

Enable customized env variables, not only tied to the values hardcoded now in reconcile. For example, for MODEL_PATH env variable, different models may require different values for it than its pure top level path. 

Instead of appending env variable directly into container, switch to use the logic appending if not exists, so that:
1. Enable customizations for env variables in runtime;
2. No error out caused by duplicate env variable if customers add them in runtime.

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/components/base.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/base.go
@@ -158,8 +158,8 @@ func UpdateEnvVariables(b *BaseComponentFields, isvc *v1beta1.InferenceService, 
 		// Base model serving - add MODEL_PATH env variable if necessary
 		if isvcutils.IsOriginalModelVolumeMountNecessary(objectMeta.Annotations) {
 			if b.BaseModel != nil && b.BaseModel.Storage != nil && b.BaseModel.Storage.Path != nil {
-				b.Log.Info("Base model serving - adding MODEL_PATH env variable", "inference service", isvc.Name, "namespace", isvc.Namespace)
-				isvcutils.AppendEnvVars(container, &[]corev1.EnvVar{
+				b.Log.Info("Base model serving - adding MODEL_PATH env variable if not provided", "inference service", isvc.Name, "namespace", isvc.Namespace)
+				isvcutils.AppendEnvVarsIfNotExist(container, &[]corev1.EnvVar{
 					{Name: constants.ModelPathEnvVarKey, Value: *b.BaseModel.Storage.Path},
 				})
 			}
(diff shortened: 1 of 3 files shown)


pkg/controller/v1beta1/inferenceservice/components/base.go
```diff
@@ -176,15 +175,13 @@ func UpdateEnvVariables(b *BaseComponentFields, isvc *v1beta1.InferenceService,
 						objectMeta.Annotations[constants.FineTunedAdapterInjectionKey],
 					),
 				})
-				isvcutils.AppendEnvVars(container, &[]corev1.EnvVar{
```
slin1237, line 179: should this function be fixed?
instead of writing a new one?

beiguo218: Adding a new function here since I am thinking if `AppendEnvVars` can serve a different purpose, to be used to do batch appending when we are certain no duplicates in current container.

beiguo218: While I gave a further thought, probably I did overthinking and created some unnecessary requirements there..

beiguo218: Let me update to fix the current function, so that it can use it for both batch and single appending and also check for duplicates in the meanwhile.

pkg/controller/v1beta1/inferenceservice/components/base.go
```diff
@@ -158,10 +158,9 @@ func UpdateEnvVariables(b *BaseComponentFields, isvc *v1beta1.InferenceService,
 		// Base model serving - add MODEL_PATH env variable if necessary
 		if isvcutils.IsOriginalModelVolumeMountNecessary(objectMeta.Annotations) {
 			if b.BaseModel != nil && b.BaseModel.Storage != nil && b.BaseModel.Storage.Path != nil {
-				b.Log.Info("Base model serving - adding MODEL_PATH env variable", "inference service", isvc.Name, "namespace", isvc.Namespace)
-				isvcutils.AppendEnvVars(container, &[]corev1.EnvVar{
-					{Name: constants.ModelPathEnvVarKey, Value: *b.BaseModel.Storage.Path},
-				})
+				b.Log.Info("Base model serving - adding MODEL_PATH env variable if it does not provided", "inference service", isvc.Name, "namespace", isvc.Namespace)
```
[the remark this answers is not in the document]
beiguo218: resolved

pkg/controller/v1beta1/inferenceservice/components/base.go
beiguo218: Resolved

pallasathena92: didn't find new change

PR #333, opened 2025-11-26
[Core] Enable Multi-Node Ray Cluster VLLM Deployment in inference service

beiguo218:
## What type of PR is this?
/kind improvement

Enables the support for deploying VLLM inference services across multiple nodes using Ray Cluster.

- New deployment mode: Integrated `MultiNodeRayVLLM` deployment mode into the engine/decoder reconcile logic;
- Deployment mode determination: Updated logic to detect `MultiNodeRayVLLM` via engine annotation;
- Added test cases

Tested changes in cluster - successfully hosted llama3.1 405b model using `MultiNodeRayVLLM` deployment mode.


slin1237:
> > why we wanna do this? Instead of using LWS? check our LWS docker image build It can run vLLM with LWS without ray operator
> 
> We still have live serving cases in our prod environment using this vLLMRayCluster deployment mode, want to keep it unchanged as we switch to OpenSource OME.
> 
> Switch to LWS multi-node would require to rewrite our current runtime, and also involve more tests there. May I please ask what is the concern to enable it in code? Its reconciling code is there and I just wired it as an option in engine so that it can be used.

because ray operator sucks ass
i hate that dependency

PR #387, opened 2025-12-04
[core] Skip Model Artifact Deletion Feature

### LINKED ISSUE #378: [ENHANCEMENT] annotation/label to prevent model artifact to be deleted
## What would you like to be added?
add a specific annotation or labels to cluster base model and base model resource
such that, if the annotation/label is added, when the model resource is deleted, it will not delete the underline model artifacts

## Why is this needed?
sometimes, when model is deleted, user do not like to delete the artifacts, or if the artifacts was already downloaded before the custom resource
deleting the resource should not trigger file delete

truddy0:
Introduce the logic of skipping the model artifact file on the node if the target label exists

## What this PR does
Add the logic of skipping model artifact file on the tnode if ```models.ome/reserve-model-artifact``` label exists on the model custom resource

## Why we need it

[Fixes #378]

## How to test

new method covered by unite test

## Checklist

- [ x] Tests added/updated (if applicable)

### DIFF
--- a/pkg/modelagent/gopher.go
+++ b/pkg/modelagent/gopher.go
@@ -884,6 +884,10 @@ func (s *Gopher) verifyDownloadedFiles(ociOSDataStore *ociobjectstore.OCIOSDataS
 }
 
 func (s *Gopher) deleteModel(destPath string, task *GopherTask) error {
+	if s.isReservingModelArtifact(task) {
+		return nil
+	}
+
 	startTime := time.Now()
 
 	err := os.RemoveAll(destPath)
@@ -902,6 +906,20 @@ func (s *Gopher) deleteModel(destPath string, task *GopherTask) error {
 	}
 
 	return err
+}
+
+// isReservingModelArtifact reports whether the model's artifact directory should be preserved during deletion.
+// It returns true when the BaseModel has the label constants.ReserveModelArtifact set to "true" (case-insensitive).
+// In all other cases, it logs that the artifact will be deleted and returns false.
+// Note: Only BaseModel labels are considered by this helper; ClusterBaseModel labels are not checked here.
+func (s *Gopher) isReservingModelArtifact(task *GopherTask) bool {
+	if val, exists := task.BaseModel.Labels[constants.ReserveModelArtifact]; exists && strings.EqualFold(val, "true") {
+		s.logger.Infof("Model artifact will be reserved as model has matched label")
+		return true
+	} else {
+		s.logger.Infof("Model artifact will be deleted")
+		return false
+	}
 }
 
 // processHuggingFaceModel handles downloading models from Hugging Face Hub.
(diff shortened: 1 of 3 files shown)


pkg/modelagent/gopher.go
```diff
@@ -904,6 +908,20 @@ func (s *Gopher) deleteModel(destPath string, task *GopherTask) error {
 	return err
 }
 
+// isReservingModelArtifact reports whether the model's artifact directory should be preserved during deletion.
+// It returns true when the BaseModel has the label constants.ReserveModelArtifact set to "true" (case-insensitive).
+// In all other cases, it logs that the artifact will be deleted and returns false.
+// Note: Only BaseModel labels are considered by this helper; ClusterBaseModel labels are not checked here.
+func (s *Gopher) isReservingModelArtifact(task *GopherTask) bool {
```
slin1237, line 915: can we use annotation instead

truddy0: Sure, may I ask why annotation is more preferred? Per my understanding, annotations can carry more data, but they cannot be used to select groups as labels do.  So I guess we don't have a use case for selecting models that reserve artifacts vs. delete artifacts?

pkg/modelagent/gopher.go
```diff
@@ -884,6 +884,10 @@ func (s *Gopher) verifyDownloadedFiles(ociOSDataStore *ociobjectstore.OCIOSDataS
 }
 
 func (s *Gopher) deleteModel(destPath string, task *GopherTask) error {
+	if s.isReservingModelArtifact(task) {
```
pallasathena92, line 887: What is the purpose of keeping these artifacts? We’ve already deleted the model CR, and retaining additional models may cause memory issues on the node.
Also, we have skip deletion because we have annotation or lables. how about adding it after this https://github.com/sgl-project/ome/blob/bcf6e4a52ee3fa95a52446489ef5b3cf3ba07c9d/pkg/modelagent/gopher.go#L439
it will easy to understand why we skip the deletion?

slin1237 (approved):

pkg/modelagent/gopher.go
truddy0: > What is the purpose of keeping these artifacts? We’ve already deleted the model CR, and retaining additional models may cause memory issues on the node.

Because some customers do not want to re-download the model or the model is pre-existing, they do not want this file to be removed. 

>  Also, we have skip deletion because we have annotation or lables. how about adding it after this 

OK. I think what you want is to handle 2 cases that will skip the artifact deletion together in this else-if block: 1) the destination path is still referenced by other model 2) this new ask - target annotation exists.

I think it is a good idea, but not ideal as there is another case where deletion should be skip when the error is popped when determining whether the path is referenced by other model. Also, if the logic is wrapped in the ```deleteModel``` func, it could be re-used in 2 storage type: StorageTypeOCI and StorageTypeHuggingFace without touching if-else blocks in gopher file. It will be clearer.

Thank you for your review and comment, and please feel free to let me know your thoughts to my reply.

pallasathena92: The reason I proposed adding it afther the else if check for “the destination path is still referenced by another model” is:

1. It’s a similar condition — normally we delete the model when we delete the model CR, but right now we explicitly skip that for some specially required.

2. Adding it here with a clear log message makes it more obvious to others why we skip deletion in this case.

PR #410, opened 2025-12-10
[Misc] Add support to more patterns of model versions

beiguo218:
## What this PR does
This PR adds the supports to the below pattern of model version:
1. model version with prefix `v` - TensorRT-LLM has versions beginning with `v`;
2. model version with single value or 2-part value (e.g. 1, 1.0) - onnx model format version can be this case.

## Changes
1. Major Version Prefix Support
 - Added `MajorPrefix` field to the `Version` struct to track version prefixes (currently supports "v" prefix)
 - Version comparison enforces prefix matching - versions with different prefixes (e.g., "v1.0.0" vs "1.0.0") are considered incompatible

2. Version Precision Support
 - Added `Precision` field to the Version struct to track the number of version parts specified:
    - Precision 1: Major version only (e.g., "1", "v1")
    - Precision 2: Major and minor versions (e.g., "1.12", "v1.12")
    - Precision 3: Major, minor, and patch versions (e.g., "1.12.3", "v1.12.3") 
 - Version comparison enforces precision matching - versions with different precisions (e.g., "1.0" vs "1.0.0") are considered incompatible

3. Added unit tests

## Checklist


pkg/modelver/util.go
```diff
@@ -13,21 +13,34 @@ const (
...
+	Precision   int // Number of version parts specified (1=major, 2=major.minor, 3=major.minor.patch)
 }
 
 // Parse parses version strings like:
 //   - 4.51.3-SAM-HQ-preview
 //   - 4.43.0.dev0
 //   - 4.43.0+build
 //   - 0.6.0
+//   - v0.8.0 (only lowercase "v" is supported as major version prefix)
+//   - 1 (single major version)
```
pallasathena92, line 40: 🥲

pkg/modelver/util.go
```diff
@@ -36,28 +49,45 @@ func Parse(s string) (Version, error) {
...
-	major, err := parseNumeric(parts[0], "major")
+	major, err := parseNumeric(majorStr, "major")
 	if err != nil {
 		return Version{}, err
 	}
 
 	// Minor
-	minor, err := parseNumeric(parts[1], "minor")
+	minorStr := "0"
+	if len(parts) > 1 {
```
pallasathena92: can use precision directly

pkg/modelver/util.go
```diff
@@ -36,28 +49,45 @@ func Parse(s string) (Version, error) {
...
-	v := Version{Major: major, Minor: minor}
+	v := Version{Major: major, MajorPrefix: majorPrefix, Minor: minor, Precision: precision}
 
+	// Patch
 	var (
 		build, prerelease, dev []string
 	)
-	patchStr := parts[2]
+	patchStr := "0"
+	if len(parts) > 2 {
```
pallasathena92: same here

pkg/modelver/util.go
```diff
@@ -36,28 +49,45 @@ func Parse(s string) (Version, error) {
...
 
-	v := Version{Major: major, Minor: minor}
+	v := Version{Major: major, MajorPrefix: majorPrefix, Minor: minor, Precision: precision}
 
+	// Patch
 	var (
 		build, prerelease, dev []string
 	)
-	patchStr := parts[2]
+	patchStr := "0"
```
pallasathena92, line 87: it might be 1.2.2+build. for this, the patchStr should be 2+build.
if patchStr doesn't exist, patchStr := parts[2] set it as 0. we don't need to check len(parts) here?

pkg/modelver/util.go
beiguo218: ¯\_(ツ)_/¯

pkg/modelver/util.go
beiguo218: If patchString does not exist, when doing parts[2], will get index out of range error.

slin1237 (approved):

PR #411, opened 2025-12-10
[API] Add model capabilities for the supported endpoints

kunzheng926:
## What this PR does:

- Adds an explicit endpoints field and ModelEndpoint enum so models declare which OpenAI v1 endpoints they support.
- Keeps existing capabilities but stops overloading them for endpoint coverage; new constants cover images, audio, and realtime.
- Plumbs endpoints through model metadata/config so agents and consumers can see supported endpoints (e.g., gpt-image-1 exposes images/generations and images/edits; grok-2-image exposes only images/generations).

## Why we need it:

- To safely enable newer endpoints (image generation/edits, audio speech/transcription/translation, realtime), callers must know a model’s supported endpoints up front.

Fixes #

## How to test

N/A

## Checklist

### DIFF
--- a/pkg/apis/ome/v1beta1/model.go
+++ b/pkg/apis/ome/v1beta1/model.go
@@ -134,10 +134,10 @@ type BaseModelSpec struct {
 	// +optional
 	ModelCapabilities []string `json:"modelCapabilities,omitempty"`
 
-	// Endpoints supported by the model, e.g., "OPENAI_V1_CHAT_COMPLETIONS"
+	// API capabilities supported by the model, e.g., "OPENAI_V1_CHAT_COMPLETIONS"
 	// +listType=atomic
 	// +optional
-	Endpoints []ModelEndpoint `json:"endpoints,omitempty"`
+	ApiCapabilities []ModelAPICapability `json:"apiCapabilities,omitempty"`
 
 	// Configuration of the model, stored as generic JSON for flexibility.
 	// +optional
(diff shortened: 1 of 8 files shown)


pkg/apis/ome/v1beta1/model.go
```diff
@@ -219,6 +219,17 @@ const (
 	ModelCapabilityTextToText      ModelCapability = "TEXT_TO_TEXT"
 	ModelCapabilityImageTextToText ModelCapability = "IMAGE_TEXT_TO_TEXT"
 	ModelCapabilityUnknown         ModelCapability = ""
+
+	// Endpoints capabilities
+	ModelCapability_OpenAI_V1_CHAT_COMPLETIONS     ModelCapability = "OPENAI_V1_CHAT_COMPLETIONS"
```
pallasathena92: what is the diffference between ModelCapability_OpenAI_V1_CHAT_COMPLETIONS and ModelCapabilityChat?

pkg/apis/ome/v1beta1/model.go
```diff
@@ -219,6 +219,17 @@ const (
...
+	ModelCapability_OpenAI_V1_RESPONSES            ModelCapability = "OPENAI_V1_RESPONSES"
```
pallasathena92: what if the model support v1-chat-completions and v1-responses?

pkg/apis/ome/v1beta1/model.go
kunzheng926: `ModelCapability_OpenAI_V1_CHAT_COMPLETIONS` indicates that the model runtime explicitly supports the OpenAI v1 /chat/completions API. This capability is tied to a specific provider API.

`ModelCapabilityChat`, on the other hand, is a legacy, generic chat capability. It does not specify which provider API the model supports (Cohere, OpenAI, Gemini, etc.)

pkg/apis/ome/v1beta1/model.go
kunzheng926: Model capabilities are represented as an array, so a model can include multiple capability values to indicate support for both endpoints.

pkg/apis/ome/v1beta1/model.go
```diff
@@ -134,6 +134,11 @@ type BaseModelSpec struct {
 	// +optional
 	ModelCapabilities []string `json:"modelCapabilities,omitempty"`
 
+	// Endpoints supported by the model, e.g., "OPENAI_V1_CHAT_COMPLETIONS"
+	// +listType=atomic
+	// +optional
+	Endpoints []ModelEndpoint `json:"endpoints,omitempty"`
```
pallasathena92: we used Endpoint for other meaning.
The ModelEndpoint is a little confusing. 
would you mind change a name?
how about supportedApis, apiCapabilities or others?

kunzheng926: thanks. changed to ModelAPICapability and apiCapabilities.

PR #419, opened 2025-12-11
[Bugs]deploymentname large than label 63 limits

pallasathena92:
## What this PR does
1. Fixed bug where long InferenceService names cause deployment failures due to Kubernetes' 63-character limit on labels and DNS names
2. Removed legacy DAC migration code that appended "-new" suffix to deployment names

## Why we need it
1. Added truncateWithHashTweaks() - truncates strings to max length using format {hash_prefix}-{suffix}. And ensures first character is always a letter (required for DNS-1035)
In order to keep minial code change only update labelsUpdated reconcilers to use truncation:
deployment_reconciler.go - truncate "app" label selector
service_reconciler.go - truncate service name and selector
pdb_reconciler.go - truncate PDB selector
domain_service.go - use new truncation function
hpa_reconciler.go - simplified to use componentMeta.Name directly
keda_reconciler.go - simplified to use componentMeta.Name directly
2. Removed legacy code:
we add -new in deployment to force create a new deployment during migration to add some immutable annotation and labels. After we have engine, decoder, router in deployment name, it will have the same effect. the "-new" is necessary and useless.
Removed updateDeploymentName() function that appended "-new" suffix for DAC/Kueue migration
Removed redundant getDeploymentName() calls from HPA and KEDA reconcilers.

## Checklist

- [ x] Tests added/updated (if applicable)
- [x ] `make test` passes locally

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -688,9 +688,8 @@ func DefaultRayHeadServiceName(name string, index int) string {
 	return rayutils.CheckName(fmt.Sprintf("%s-%d", name, index))
 }
 
-func TruncateDomainName(name string, maxLength int) string {
-	return truncateWithHash(name, maxLength)
-}
+// label has 63 characters limit
+// we will have inferenceservicename-engine, inferenceservicename
 
 // Kubernetes naming constraints
 const (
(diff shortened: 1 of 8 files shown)


pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
```diff
@@ -48,28 +48,14 @@ func createHPA(componentMeta metav1.ObjectMeta,
...
-	 *     it cannot directly update the existing old OME deployment resource due to above point #1;
-	 *  Note: Only need to adopt a new deployment name when it comes to migrate old OME DAC inference service, no need to do
-	 *        this for below:
-	 *     1). on-demand model serving;
-	 *     2). DAC inference service deployment from new OME with Volcano reconciled; (Out of scope, will handle its migration
-	 *         separately)
-	 *  when we create an inference service with Kueue enabled, we need to use inferencerservice name + "-new" as the deployment name
-	 *  for hpa scale target ref.
-	 */
-	deploymentName := getDeploymentName(componentMeta)
```
beiguo218, line 64: We can keep using `getDeploymentName` function since it is just returning `componentMeta.Name`. 
We can just remove the comment here to avoid the confusion.

pkg/constants/constants.go
```diff
@@ -688,9 +688,8 @@ func DefaultRayHeadServiceName(name string, index int) string {
 	return rayutils.CheckName(fmt.Sprintf("%s-%d", name, index))
 }
 
-func TruncateDomainName(name string, maxLength int) string {
-	return truncateWithHash(name, maxLength)
-}
+// label has 63 characters limit
```
beiguo218, line 691: nit: what this part of comment is for? probably either remove it or move it to correct place

pkg/controller/v1beta1/inferenceservice/reconcilers/keda/keda_reconciler.go
```diff
@@ -63,20 +63,6 @@ func createScaledObject(
...
-	 *     it cannot directly update the existing old OME deployment resource due to above point #1;
-	 *  Note: Only need to adopt a new deployment name when it comes to migrate old OME DAC inference service, no need to do
-	 *        this for below:
-	 *     1). on-demand model serving;
-	 *     2). DAC inference service deployment from new OME with Volcano reconciled; (Out of scope, will handle its migration
-	 *         separately)
-	 *  when we create an inference service with Kueue enabled, we need to use inferencerservice name + "-new" as the deployment name
-	 *  for keda scale target ref.
-	 */
-	deploymentName := getDeploymentName(componentMeta)
```
beiguo218, line 79: Remove unused `getDeploymentName` in keda_reconciler

pkg/controller/v1beta1/inferenceservice/reconcilers/hpa/hpa_reconciler.go
beiguo218: or just remove `getDeploymentName` function in hpa_reconcile.go

Issue #417, opened 2025-12-11
[BUG] Annotations from engine/router on pods not passed down to pod spec reconcile


beiguo218:
## What happened?

With changes in PR: https://github.com/sgl-project/ome/pull/391, we found the metadatas (annotations/labels) from engine/router on pods not being passed down to the pod spec reconcile logic so causing wrong determinations in pod spec reconcile logic where we need to rely on the values of those annotations/labels.

The wrong determinations we found so far in the pod spec reconcile logic (caused by https://github.com/sgl-project/ome/pull/391)

1. wrong env variable added;
2. wrong volume mount added;
3. wrong volume added.

## What did you expect to happen?
It is okay to make changes to avoid unnecessary annotations/labels added to the unwanted components. The expectation is for the places where the logic needs to rely on those metadatas (annotations/labels from engine/router on pods), we need to pass down the correct metadatas, to make sure the existing reconcile logic won't break.

The expectation for such changes:
1. Make sure to mention explicitly in PR description what components you target to change to have different metadata;
2. Make sure you still pass the original metadata when there is a need to rely on its value.

Under the case your goal has conflicts with current reconciling logic, you should open an issue.

## How can we reproduce it (as minimally and precisely as possible)?

Under the case where the runtime spec is with an annotation `ome.io/inject-model-init: 'true'` under `engineConfig` which we use to inject the model init for model enigma usage, with changes from https://github.com/sgl-project/ome/pull/391, this annotations from engine won't be passed down to pod spec reconcile logic so causing unwanted env variable (MODEL_PATH) and unwanted volume mount added to our serving container.

Example runtime with `ome.io/inject-model-init: 'true'` annotation:
```
apiVersion: ome.io/v1beta1
kind: ClusterServingRuntime
metadata:
  annotations:
    meta.helm.sh/release-name: ome-model-serving
    meta.helm.sh/release-namespace: ome
  labels:
    app.kubernetes.io/managed-by: Helm
  name: sample-runtime
spec:
  engineConfig:
    affinity:
      nodeAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
          nodeSelectorTerms:
            - matchExpressions:
                - key: node.kubernetes.io/instance-type
                  operator: In
                  values:
                    - BM.GPU.H100.8
    annotations:
      ome.io/inject-model-init: 'true'
    labels:
      logging-forward: enabled
    runner:
      image: sample-image:sample-tag
      ports:
        - containerPort: 8080
          name: http1
[...]

PR #501, opened 2026-01-11
Do not add grafana related annotations to k8s service

Kangyan-Zhou:
## What this PR does

retry https://github.com/sgl-project/ome/pull/391. Right now the grafana annotations on the services will cause grafana to scrape the service as well and lead to wrong metrics.

## Why we need it

Fixes https://github.com/sgl-project/ome/issues/417

## How to test

Tested locally and the annotations on the service are gone

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/service/service_reconciler_test.go
@@ -12,8 +12,8 @@ import (
 
 func TestBuildServiceFiltersAnnotations(t *testing.T) {
 	scenarios := map[string]struct {
-		componentMeta        metav1.ObjectMeta
-		expectedAnnotations  map[string]string
+		componentMeta         metav1.ObjectMeta
+		expectedAnnotations   map[string]string
 		unexpectedAnnotations []string
 	}{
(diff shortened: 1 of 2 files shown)


pkg/constants/constants.go
```diff
@@ -397,6 +397,26 @@ var (
...
+	// PodOnlyAnnotationPrefixes contains annotation prefixes (or exact keys) that should only
+	// be applied to Pods, not to Services. These are typically used for pod-level configurations
+	// like monitoring, networking interfaces, container injection, and other pod-specific settings.
+	//
+	// Entries can be either:
+	// - Prefix patterns ending with "/" (e.g., "k8s.grafana.com/") - matches all annotations with this prefix
+	// - Exact annotation keys (e.g., ModelInitInjectionKey) - matches only that specific annotation
+	//
+	// Both styles work correctly because IsPrefixSupported uses strings.HasPrefix for matching.
+	PodOnlyAnnotationPrefixes = []string{
```
fzyzcjy, line 410: nit: a bit worried hardcoding this will be ok or not

Kangyan-Zhou: I couldn't think of a good way to dynamically include all labels, @slin1237 please suggest if you have a better way to avoid hardcoding here

PR #505, opened 2026-01-16
[OEP] Add OEP-0006: Workload Policy Layer and Unified Workload Manage

bcfre:
## What this PR does

This PR introduces OEP-0006, which proposes a Workload Policy Layer architecture for unified workload management in OME.

## Why we need it

This OEP introduces a new abstraction layer for workload management that:
- Decouples workload orchestration logic from role reconciliation logic
- Supports both SingleComponent (independent deployment) and All-in-One (RoleBasedGroup) deployment modes
- Provides an extensible framework for future workload strategies

Before introducing the workload policy layer, OME's workload deployment logic was tightly coupled within each role's Reconciler, with each component independently creating Deployment, Service, Knative Service, and other resources. This architecture has the following issues:

1. **Insufficient Operability**: Does not support in-place image upgrades: existing workloads must rebuild Pods when images are updated and lacks specified warm-up control capabilities, causing service interruption or delays and slowing overall release efficiency.

2. **Weak Multi-Role Coordination**: Lacks unified cross-role orchestration: each role reconciles independently, unable to perceive each other's versions, states, or dependencies. During scaling or upgrades, version consistency or proportional coordination between roles cannot be guaranteed (such as "upgrade A then B", "A:B = 2:1" strategies), easily causing compatibility issues or service anomalies.

As distributed high-performance inference increases demands for coordination between different inference roles, we need to introduce All-in-One workload (RoleBasedGroup) support in OME to fine-tune all roles from a global perspective.

## Checklist

- [x] Docs updated (if applicable)

### DIFF
--- a/oeps/0006-unified-workload-management/README.md
+++ b/oeps/0006-unified-workload-management/README.md
@@ -1,0 +1,641 @@
+# OEP-0006: Workload Policy Layer and Unified Workload Management
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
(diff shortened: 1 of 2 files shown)


cheyang:
@slin1237 Please take a look. Thanks.

slin1237 (approved):

PR #519, opened 2026-02-05
realtime model capability

upfixer:
## What this PR does

Add model capability REALTIME

Fixes #
support realtime models, api capability already has realtime

## Checklist


pkg/apis/ome/v1beta1/model.go
```diff
@@ -292,6 +292,7 @@ const (
 	ModelCapabilityTextToSpeech     ModelCapability = "TEXT_TO_SPEECH"
 	ModelCapabilitySpeechToText     ModelCapability = "SPEECH_TO_TEXT"
 	ModelCapabilityAudioTranslation ModelCapability = "AUDIO_TRANSLATION"
+	ModelCapabilityRealtime         ModelCapability = "REALTIME"
```
beiguo218, line 295: You don’t need to add it here — you can just use it in your model CR.

beiguo218: Feel like it is something internal, so that's why I push back.
In order to use it, you can directly use it in model CR, we do not enforce that it has to be ModelCapability enum.

PR #529, opened 2026-02-27
[Helm] Add OCI Compute Shape B200 support in model agent

heymrbox:
## What this PR does

Add OCI Compute Shape B200, GB200 and GB300 support in model agent

## Why we need it

To support the new shapes in OCI

Fixes #

## How to test

make test

## Checklist

### DIFF
--- a/pkg/utils/instance_type_util.go
+++ b/pkg/utils/instance_type_util.go
@@ -26,6 +26,7 @@ var defaultInstanceTypeMap = map[string]string{
 	"BM.GPU.H100-NC.8": "H100",
 	"BM.GPU.H200.8":    "H200",
 	"BM.GPU.H200-NC.8": "H200",
+	"BM.GPU.B200.8":    "B200",
 
 	// AWS instance types
 	"p5.48xlarge": "H100",
--- a/charts/ome-resources/templates/model-agent-daemonset/configmap.yaml
+++ b/charts/ome-resources/templates/model-agent-daemonset/configmap.yaml
@@ -21,6 +21,7 @@ data:
       "BM.GPU.H100-NC.8": "H100",
       "BM.GPU.H200.8": "H200",
       "BM.GPU.H200-NC.8": "H200",
+      "BM.GPU.B200.8": "B200",
       "p5.48xlarge": "H100",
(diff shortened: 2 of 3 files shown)


pkg/utils/instance_type_util.go
```diff
@@ -26,6 +26,9 @@ var defaultInstanceTypeMap = map[string]string{
 	"BM.GPU.H100-NC.8": "H100",
 	"BM.GPU.H200.8":    "H200",
 	"BM.GPU.H200-NC.8": "H200",
+	"BM.GPU.B200.8":    "B200",
+	"BM.GPU.GB200.4":   "GB200",
+	"BM.GPU.GB300.4":   "GB300",
```
beiguo218: In terms of how this map is used, it is actually for models with tensorrtllm model format to be parsed correctly. Are we sure for models with tensorrtllm model format, in terms of BM.GPU.GB200.4 and BM.GPU.GB300.4, `GB200` and `GB300` are used in model path? Asked it since for both BM.GPU.GB200.4 and BM.GPU.GB300.4, based on what I know, the underlying shape actually be `B200` and `B300`.

go.sum
```diff
@@ -1340,7 +1340,9 @@ k8s.io/component-helpers v0.31.4/go.mod h1:Ddq5GYRK/1uNoPNgJh9N5osPutvBweQEcIG6b
 k8s.io/controller-manager v0.30.12/go.mod h1:vbZ0OY4XMFGJDl9lTaPGc+BDocukZzMH41iNVc5BcY8=
 k8s.io/csi-translation-lib v0.30.12/go.mod h1:5lNKaS8VpgmTpH6rPAdwEbQ9HvCnpOGaqpP7OAUFj/I=
 k8s.io/dynamic-resource-allocation v0.30.12/go.mod h1:+1QBI9MRebuOMtLAqCK/RB+oKCh9f+2EhAgkEOV7Fz8=
+k8s.io/gengo v0.0.0-20230829151522-9cce18d56c01 h1:pWEwq4Asjm4vjW7vcsmijwBhOr1/shsbSYiWXmNGlks=
```
beiguo218: Can we please cleanup these changes in go.sum

PR #533, opened 2026-03-03
(feat)(fix) Fix Qwen quantization parsing and add Qwen 3.5 model config tests

YouNeedCryDear:
## What this PR does

- Fixes quantization config parsing for Qwen model configurations.
- Adds Qwen 3.5 model config parsing support and fixtures.
- Expands unit test coverage across Qwen parser variants, including FP8 cases.

## Why we need it

Current parsing for newer Qwen model quantization config is incomplete/inaccurate in some cases, which can lead to incorrect model metadata extraction. These updates improve correctness and add regression coverage.

Fixes #

## How to test

- `go test ./pkg/hfutil/modelconfig/...`

## Checklist

### DIFF
--- a/pkg/hfutil/modelconfig/qwen2_test.go
+++ b/pkg/hfutil/modelconfig/qwen2_test.go
@@ -109,12 +109,12 @@ func TestQwen2QuantizationConfig(t *testing.T) {
 		t.Fatalf("Failed to unmarshal Qwen2 FP8 config: %v", err)
 	}
 
-	if config.GetQuantizationType() != "fp8" {
-		t.Errorf("Expected quantization type 'fp8', but got '%s'", config.GetQuantizationType())
-	}
-
 	if config.QuantizationConfig == nil {
 		t.Fatal("Expected QuantizationConfig to be non-nil")
+	}
+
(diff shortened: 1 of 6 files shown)


pkg/hfutil/modelconfig/qwen2_test.go
```diff
@@ -71,9 +72,52 @@ func TestQwen2Config(t *testing.T) {
...
+	config := &Qwen2Config{}
+	if err := json.Unmarshal(jsonData, config); err != nil {
+		t.Fatalf("Failed to unmarshal Qwen2 FP8 config: %v", err)
+	}
+
+	if config.GetQuantizationType() != "fp8" {
+		t.Errorf("Expected quantization type 'fp8', but got '%s'", config.GetQuantizationType())
+	}
+
+	if config.QuantizationConfig == nil {
```
pallasathena92, line 112: nit. if QuantizationConfig == nil, the config.GenQuantizationType() != "fp8" has already failed.

YouNeedCryDear: reordered the checking sequence.

PR #555, opened 2026-04-01
feat: optional InferenceService annotation to skip model-ready nodeSelector (Karpenter)

ossianpe:
## Problem
Engine/decoder pods get a required `nodeSelector` of the form `models.ome.io/clusterbasemodel.<name>=Ready`. The model-agent applies that label only after weights are on disk. Autoscalers such as Karpenter may not provision nodes when the pod requires labels that no NodePool advertises yet, causing a scheduling deadlock on cold GPU pools.

## Solution
- New annotation on **InferenceService**: `ome.io/skip-model-ready-node-selector: "true"` (default behavior unchanged when absent).
- When set, **omit** the model-ready `nodeSelector`; accelerator/runtime merged selectors still apply.
- **BenchmarkJob**: if the referenced InferenceService has the annotation, the benchmark pod skips the same selector for consistency.

## Implementation
- `constants.SkipModelReadyNodeSelectorAnnotationKey`
- `IsSkipModelReadyNodeSelector()` + gate in `UpdatePodSpecNodeSelector` (alongside existing merged-weights skip).
- Unit tests for helper and `UpdatePodSpecNodeSelector`.
- Docs: `charts/ome-resources/README.md` (Autoscaling / Karpenter section).

## Trade-off
Pods may land on a GPU node before the model exists on hostPath; operators rely on model-agent download, restarts, or other guards (see README).

## Build ergonomics (Makefile)
Optional `BASE_IMAGE=ubuntu:24.04` for `linux/amd64` image builds on Apple Silicon (Oracle Linux 10 runtime can fail under QEMU with `CPU does not support x86-64-v3`).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
+++ b/pkg/controller/v1beta1/inferenceservice/utils/utils_test.go
@@ -1870,6 +1870,25 @@ func TestGetTargetServicePort_ServiceNameResolution(t *testing.T) {
 	}
 }
 
+func TestIsSkipModelReadyNodeSelector(t *testing.T) {
+	tests := []struct {
+		name        string
+		annotations map[string]string
+		want        bool
+	}{
+		{name: "nil annotations", annotations: nil, want: false},
+		{name: "empty annotations", annotations: map[string]string{}, want: false},
+		{name: "true", annotations: map[string]string{constants.SkipModelReadyNodeSelectorAnnotationKey: "true"}, want: true},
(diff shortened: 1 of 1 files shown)


ossianpe:
We decided we want to do more reviewing/testing before submitting this for review. Sorry for the noise, but take it as a heads up for an incoming change request for kubernetes autoscalers.

PR #574, opened 2026-04-09
[Misc] add Whisper config parser

OpheliaLjh:
## Motivation

Adds a config parser for OpenAI Whisper speech recognition models (e.g. `openai/whisper-large-v3`, `openai/whisper-large-v3-turbo`) to `pkg/hfutil/modelconfig`. Without this, `ModelAgent` cannot inspect Whisper checkpoints and serving runtimes that rely on model-config dispatch (size range, architecture match) cannot schedule Whisper models.

This is blocking guardrail STT / audio transcription work that depends on serving Whisper under OME.

## Modifications

- Add `WhisperConfig` struct implementing the `HuggingFaceModel` interface.
- Model Whisper as an encoder-decoder transformer with separate encoder/decoder dimensions (`encoder_layers`, `decoder_layers`, `encoder_attention_heads`, `decoder_attention_heads`, `encoder_ffn_dim`, `decoder_ffn_dim`, `d_model`) rather than the single `num_hidden_layers` / `hidden_size` pair used by causal LMs.
- Expose audio / position limits (`num_mel_bins`, `max_source_positions`, `max_target_positions`).
- `GetContextLength()` returns `max_target_positions` (the decoder token budget, 448 on every published OpenAI checkpoint), which matches what callers use when sizing requests against the OpenAI-compatible transcription API.
- `GetParameterCount()` first reads from co-located safetensors files via `FindAndParseSafetensors`, then falls back to a lookup table for the well-known OpenAI Whisper sizes (`tiny`, `base`, `small`, `medium`, `large` / `large-v2` / `large-v3`, `large-v3-turbo`).
- Register `"whisper"` in the model loader registry via `init()` so `LoadModelConfig` dispatches correctly.
- Add `whisper_test.go` with a direct-type test and a dispatcher round-trip test.
- Add `testdata/whisper_large_v3_turbo.json` — a verbatim copy of the published HuggingFace `config.json` for `openai/whisper-large-v3-turbo`.

## Checklist

- [x] Format your code with [gofumpt](https://github.com/mvdan/gofumpt) and [golines](https://github.com/segmentio/golines) (`make format`).
- [x] Add unit tests as outlined in the [Modifications](#modifications) section.
- [x] Update documentation / examples if applicable.


pkg/hfutil/modelconfig/whisper.go
```diff
@@ -0,0 +1,203 @@
...
+		return 769_000_000 // whisper-medium (~769M)
+	case c.EncoderLayers == 12 && c.DecoderLayers == 12 && c.DModel == 768:
+		return 244_000_000 // whisper-small (~244M)
+	case c.EncoderLayers == 6 && c.DecoderLayers == 6 && c.DModel == 512:
+		return 74_000_000 // whisper-base (~74M)
+	case c.EncoderLayers == 4 && c.DecoderLayers == 4 && c.DModel == 384:
+		return 39_000_000 // whisper-tiny (~39M)
+	}
+
+	return 0
```
pallasathena92, line 143: let's use a fallback method rather than return 0 here. Detail is in other comment.

pallasathena92 (commented):
## Suggestion: add `estimateWhisperParamCount` as a final fallback

Right now `GetParameterCount()` returns `0` when safetensors are unavailable and the model doesn't match a known checkpoint. Other model configs in the repo (e.g. `mllama.go`, `llama4.go`) use estimation helpers as a last resort. A Whisper equivalent would cover fine-tuned or future variants that don't appear in the hard-coded table.

```go
// estimateWhisperParamCount estimates the total parameter count for a Whisper
// encoder-decoder model from its architectural dimensions.
func estimateWhisperParamCount(dModel, encoderLayers, decoderLayers, encoderFfnDim, decoderFfnDim, vocabSize, numMelBins int) int64 {
	// Encoder: self-attention + FFN per layer
	encoderAttn := int64(dModel * dModel * 4 * encoderLayers)
	encoderFfn := int64(dModel * encoderFfnDim * 2 * encoderLayers)

	// Decoder: self-attention + cross-attention + FFN per layer
	decoderAttn := int64(dModel * dModel * 8 * decoderLayers) // 4 for self + 4 for cross
	decoderFfn := int64(dModel * decoderFfnDim * 2 * decoderLayers)

	// Token embedding + conv feature extractor (2 conv layers with kernel_size=3)
	embedding := int64(vocabSize * dModel)
	conv := int64(numMelBins*dModel*3 + dModel*dModel*3)

	return encoderAttn + encoderFfn + decoderAttn + decoderFfn + embedding + conv
}
```

Sanity check against known models:

| Model | Known params | Estimate |
|---|---|---|
| whisper-large-v3-turbo (enc=32, dec=4, d=1280) | 809M | ~805M |
| whisper-large-v3 (enc=32, dec=32, d=1280) | 1,550M | ~1,510M |
| whisper-tiny (enc=4, dec=4, d=384) | 39M | ~37M |

Estimates are within ~3%. Usage would replace the `return 0` at the end of `GetParameterCount()`:

```go
// Unknown Whisper variant — estimate from architecture dimensions
return estimateWhisperParamCount(
    c.DModel, c.EncoderLayers, c.DecoderLayers,
    c.EncoderFfnDim, c.DecoderFfnDim, c.VocabSize, c.NumMelBins)
```

pkg/hfutil/modelconfig/whisper.go
```diff
@@ -0,0 +1,203 @@
...
+// GetModelType returns the model type.
+func (c *WhisperConfig) GetModelType() string {
+	return c.ModelType
+}
+
+// GetContextLength returns the maximum context length.
+//
+// For Whisper this is the decoder token budget (max_target_positions, 448
+// for every published OpenAI checkpoint), which is what callers use when
+// sizing requests against the OpenAI-compatible API.
```
pallasathena92, line 173: whisper models Input and output live in completely different spaces:                                                                                                                                                                  
  - Input is audio, represented as mel spectrogram frames. 1 frame ≠ 1 token. 1500 frames ≈ 30 seconds of audio.                                                                                                         
  - Output is text tokens from a BPE vocabulary (51,866 entries). 448 tokens ≈ a few paragraphs of transcribed text.

The OpenAI transcription API (/v1/audio/transcriptions) doesn't have a `max_tokens` parameter, the caller won't use this size to set response size.

might let's use audio input size here?
// GetContextLength returns max_source_positions— the encoder's input token                                                                                                                                          
  // limit (1500).                                                                                                                                                                   
  //                                                                                                                                                                                                                     
  // Note: Whisper is an encoder-decoder model where input (audio) and output                                                                                                                                            
  // (text) are in different modalities. The encoder's input capacity is governed                                                                                                                                        
  // by max_source_positions (1500 frames ≈ 30s of audio), which is independent                                                                                                                                          
  // of this value. We return the encoder limit as it represents the audio generation capacity.

PR #576, opened 2026-04-10
[Core] Surface GPU OOM errors from pod logs in InferenceService status

### LINKED ISSUE #575: [BUG] InferenceService status misses pods for long names
## What happened?

For long InferenceService names, the controller may fail to find component pods during status propagation. The pods are actually labeled with a truncated/hashed app value to fit Kubernetes naming/label limits. This hides or delays accurate component status reporting and leaves status dependent only on higher-level deployment conditions.

## What did you expect to happen?

The controller should use the same label derivation for pod lookup as deployment/pod creation, so it can always find component pods and propagate status correctly.

## How can we reproduce it (as minimally and precisely as possible)?

1. Create an InferenceService with a long name.
2. Let it create a raw deployment engine pod.
3. Check controller logs during status propagation.
4. Compare:
    - the podLabelValue used by status lookup
    - the actual pod metadata.labels.app

Listed pods while updating component model status 
```
{
  "component": "engine",
  "podCount": 0,
  "podLabelKey": "app",
  "podLabelValue": "<full-inference-service-name>-engine"
}
```
But the actual pod label is truncated/hashed, for example:

yaml
```
labels:
  app: a5b5c2cf-jqa4tzjnvnaeioaw6ewzj5uevu2qlj6ii6vknafdarwgmfq-engine
```
So the selector does not match the pod.

## Impact
Even when pod lookup fails, the controller still updates `EngineReady` from the Deployment status earlier in the reconcile via `PropagateRawStatus()`. That path does not depend on matching pods. As a result, status falls back to generic deployment-level conditions like `MinimumReplicasUnavailable`, which makes debugging harder and hides the real runtime failure reason.

op109lvb:
## What this PR does

Polishes InferenceService failure reporting so GPU out-of-memory errors can be surfaced from crashed pod logs instead of only falling back to generic deployment availability conditions.

This PR:
1. Detects GPU OOM by inspecting termination messages and previous pod logs for crashlooping runtime containers
2. Sets a clearer component condition reason/message for GPU memory insufficiency
3. Fixes raw-deployment pod lookup for long InferenceService names by making status pod selection use the same truncated app label as deployment/pod creation
4. Preserves existing non-initializing conditions instead of overwriting them during condition initialization
5. Adds unit tests coverage

## Why we need it

Before this change, GPU OOM failures were easy to miss in InferenceService status.

Two things got in the way:

1. The controller often only surfaced generic deployment-level conditions such as MinimumReplicasUnavailable
2. For long InferenceService names, pod-based status propagation could fail entirely because the controller selected pods with the full component name while the actual pod app label was truncated/hashed
As a result, the controller could miss the crashed pod, skip previous-log inspection, and fail to surface the real runtime failure reason.

Fixes #575

## How to test

1. Deploy an InferenceService with a long name in raw deployment mode.
2. Use a model/runtime combination that causes the runtime container to crash with CUDA OOM.
3. Confirm the runtime pod enters CrashLoopBackOff.
4. Verify the InferenceService now matches the pod and surfaces a clearer condition instead of only generic deployment unavailability.

## Test result
With this change, the isvc status contains: 
```
status:
  conditions:
    - lastTransitionTime: '2026-04-10T00:32:07Z'
      message: Insufficient GPU memory to run the model container
      reason: InsufficientGPUMemory
      status: 'False'
      type: Ready
```

## Checklist

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/components/router.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/router.go
@@ -47,6 +47,8 @@ func NewRouter(
 	runtime *v1beta1.ServingRuntimeSpec,
 	runtimeName string,
 ) Component {
+	logger := ctrl.Log.WithName("RouterReconciler")
+	statusManager := status.NewStatusReconciler(clientset).WithLogger(logger.WithName("InferenceServiceStatus"))
 	base := BaseComponentFields{
 		Client:                 client,
 		Clientset:              clientset,
@@ -57,8 +59,8 @@ func NewRouter(
 		BaseModelMeta:          baseModelMeta,
 		Runtime:                runtime,
 		RuntimeName:            runtimeName,
-		StatusManager:          status.NewStatusReconciler(),
-		Log:                    ctrl.Log.WithName("RouterReconciler"),
+		StatusManager:          statusManager,
+		Log:                    logger,
 	}
 
 	return &Router{
--- a/pkg/controller/v1beta1/inferenceservice/controller.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller.go
@@ -78,6 +78,7 @@ import (
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
 // +kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=pods/log,verbs=get
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=persistentvolumes,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
@@ -619,7 +620,7 @@ func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, deployCo
 	r.ClientConfig = mgr.GetConfig()
 
 	// NEW: Initialize StatusReconciler
-	r.StatusManager = status.NewStatusReconciler()
+	r.StatusManager = status.NewStatusReconciler(r.Clientset)
 
 	// Initialize RuntimeSelector
 	r.RuntimeSelector = runtimeselector.New(mgr.GetClient())
--- a/pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
@@ -3,10 +3,13 @@ package status
 import (
 	"reflect"
 
+	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 	"knative.dev/pkg/apis"
 	knservingv1 "knative.dev/serving/pkg/apis/serving/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	lwsspec "sigs.k8s.io/lws/api/leaderworkerset/v1"
 
 	"github.com/sgl-project/ome/pkg/apis/ome/v1beta1"
@@ -20,11 +23,27 @@ const (
 )
 
 // StatusReconciler handles all status-related operations for InferenceService
-type StatusReconciler struct{}
+type StatusReconciler struct {
+	Clientset     kubernetes.Interface
+	Log           logr.Logger
+	podLogFetcher func(namespace, podName, containerName string, previous bool) (string, error)
+}
 
-// NewStatusReconciler creates a new StatusReconciler instance
-func NewStatusReconciler() *StatusReconciler {
-	return &StatusReconciler{}
+// NewStatusReconciler creates a new StatusReconciler instance.
+// Pass a nil clientset when Kubernetes log fetching is not needed.
+func NewStatusReconciler(clientset kubernetes.Interface) *StatusReconciler {
+	reconciler := &StatusReconciler{
+		Clientset: clientset,
+		Log:       logf.Log.WithName("InferenceServiceStatus"),
+	}
+	reconciler.podLogFetcher = reconciler.fetchPodLogs
+	return reconciler
+}
+
+// WithLogger overrides the default logger and returns the reconciler for chaining.
+func (sr *StatusReconciler) WithLogger(log logr.Logger) *StatusReconciler {
+	sr.Log = log
+	return sr
 }
 
 // PropagateRawStatus propagates status from raw Kubernetes deployment
@@ -147,10 +166,10 @@ func (sr *StatusReconciler) PropagateStatus(
 // PropagateModelStatus propagates model status from pod information
 func (sr *StatusReconciler) PropagateModelStatus(
 	status *v1beta1.InferenceServiceStatus,
+	component v1beta1.ComponentType,
 	statusSpec v1beta1.ComponentStatusSpec,
 	podList *v1.PodList,
 	rawDeployment bool) {
-
 	// Check at least one pod is running for the latest revision of inferenceservice
 	totalCopies := len(podList.Items)
 	if totalCopies == 0 {
@@ -177,7 +196,7 @@ func (sr *StatusReconciler) PropagateModelStatus(
 	}
 
 	// Check container statuses
-	sr.checkContainerStatuses(status, firstPod, totalCopies)
+	sr.checkContainerStatuses(status, component, firstPod, totalCopies)
 }
 
 // UpdateModelRevisionStates updates the model revision states
--- a/pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
+++ b/pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
@@ -21,7 +21,7 @@ const (
 	RoutesReadyCondition         = "RoutesReady"
 	ConfigurationsReadyCondition = "ConfigurationsReady"
 	defaultPodLogTailLines       = int64(200)
-	defaultPodLogLimitBytes      = int64(256 * 1024)
+	defaultPodLogLimitBytes      = int64(32 * 1024)
 )
 
 // StatusReconciler handles all status-related operations for InferenceService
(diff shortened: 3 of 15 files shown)


pkg/controller/v1beta1/inferenceservice/status/status_util.go
```diff
@@ -270,21 +273,23 @@ func (sr *StatusReconciler) checkContainerStatuses(status *v1beta1.InferenceServ
 				sr.UpdateModelRevisionStates(status, v1beta1.Loading, totalCopies, nil)
 				return
 			case cs.State.Terminated != nil && cs.State.Terminated.Reason == constants.StateReasonError:
-				message, exitCode, _ := sr.safeGetTerminationMessage(cs)
+				message, exitCode, _ := sr.getContainerFailureMessage(firstPod, cs)
```
pallasathena92: This means we will now hit the K8s API for pods/log on the storage initializer container whenever model download fails — but the storage initializer never raises GPU OOM (it doesn't touch the GPU). This is pure overhead and noise, and on a busy controller with many failing pods it adds real load to the API server.
do we really need this for storage initializer container?

pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
```diff
@@ -20,11 +23,28 @@ const (
...
+type StatusReconciler struct {
+	Clientset     kubernetes.Interface
+	Log           logr.Logger
+	podLogFetcher func(namespace, podName, containerName string, previous bool) (string, error)
+}
 
 // NewStatusReconciler creates a new StatusReconciler instance
-func NewStatusReconciler() *StatusReconciler {
-	return &StatusReconciler{}
+func NewStatusReconciler(clientsets ...kubernetes.Interface) *StatusReconciler {
```
pallasathena92: why we use a variadic clientsets parameter here?

pkg/controller/v1beta1/inferenceservice/controller.go
```diff
@@ -619,7 +620,7 @@ func (r *InferenceServiceReconciler) SetupWithManager(mgr ctrl.Manager, deployCo
 	r.ClientConfig = mgr.GetConfig()
 
 	// NEW: Initialize StatusReconciler
-	r.StatusManager = status.NewStatusReconciler()
+	r.StatusManager = status.NewStatusReconciler(r.Clientset)
```
pallasathena92, line 623: the statusManager in controller actually didn't need r.clientset. it does cross-component aggregation and propagation from K8s objects (PropagateRawStatus, PropagateMultiNodeStatus, PropagateStatus, PropagateMultiNodeRayVLLMStatus, PropagateCrossComponentStatus). Add r.Clientset no one function actually call it.
if you keey the signature in NewStatusReconcile() variadic. just remove it from here.
if you make the signture non-variadic. Make two constructors will be better

pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
op109lvb: Thanks for pointing it out. It was for adding more debugging logs to find out another issue and avoid nil case. Will clean it up. Thanks.

pkg/controller/v1beta1/inferenceservice/controller.go
op109lvb: fetchPodLogs in status_util calls Clientset to get pod logs: https://github.com/sgl-project/ome/blob/b24f38f803f476e764672bae57285d340093b8fd/pkg/controller/v1beta1/inferenceservice/status/status_util.go#L384

pkg/controller/v1beta1/inferenceservice/status/status_util.go
op109lvb: Thanks for pointing it out. Yes, will revert the change for init container.

pkg/controller/v1beta1/inferenceservice/status/status_util.go
```diff
@@ -324,12 +331,99 @@ func (sr *StatusReconciler) checkContainerStatuses(status *v1beta1.InferenceServ
...
+		return "", fmt.Errorf("clientset is not configured")
+	}
+
+	sr.Log.Info("Requesting pod logs",
+		"namespace", namespace,
+		"pod", podName,
+		"container", containerName,
+		"previous", previous)
+
+	req := sr.Clientset.CoreV1().Pods(namespace).GetLogs(podName, &v1.PodLogOptions{
```
pallasathena92, line 386: A vLLM/sglang runtime that crashloops for hours can produce megabytes of logs per container instance. This function reads the entire log into memory on every reconcile. With 50 failing ISVCs requeuing every 30 seconds, you're doing 50 unbounded log fetches per reconcile cycle. This will spike controller memory and put unnecessary load on kubelet's log endpoint.

Fix:
```
tailLines := int64(200)
  limitBytes := int64(256 * 1024) // 256 KB
  req := sr.Clientset.CoreV1().Pods(namespace).GetLogs(podName, &v1.PodLogOptions{
      Container:  containerName,
      Previous:   previous,
      TailLines:  &tailLines,
      LimitBytes: &limitBytes,
  })
```

pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
```diff
@@ -20,11 +23,27 @@ const (
...
+}
+
+// NewStatusReconciler creates a new StatusReconciler instance.
+// Pass a nil clientset when Kubernetes log fetching is not needed.
+func NewStatusReconciler(clientset kubernetes.Interface) *StatusReconciler {
+	reconciler := &StatusReconciler{
+		Clientset: clientset,
+		Log:       logf.Log.WithName("InferenceServiceStatus"),
+	}
+	reconciler.podLogFetcher = reconciler.fetchPodLogs
```
pallasathena92: This always set even with nil clientset. There are so many test with nil clientset. it spurious error logs in test.
for example:
test TestCheckContainerStatuses — the test cases "main container terminated with error" 
go through `getContainerFailureMessage`
pass the podLogFetcher == nil check
call fetchPodLogs
Hit clientset == nil -> return error
Log an error: "Failed to fetch pod logs for failure inspection" + "clientset is not configured"

suggestion:
if clientset != nil {
      reconciler.podLogFetcher = reconciler.fetchPodLogs
  }

pkg/controller/v1beta1/inferenceservice/components/router.go
```diff
@@ -47,6 +47,8 @@ func NewRouter(
 	runtime *v1beta1.ServingRuntimeSpec,
 	runtimeName string,
 ) Component {
+	logger := ctrl.Log.WithName("RouterReconciler")
```
pallasathena92, line 50: logr treat each WithName as a hierarchical level. 
As we have default logger in constuctor `logf.Log.WithName("InferenceServiceStatus")`
how about use `logger := ctrl.Log.WithName("RouterReconciler.InferenceServiceStatus")` to replace `logger.WithName("InferenceServiceStatus")` in next line.
which is simpler and does the same thing

pkg/controller/v1beta1/inferenceservice/status/status_util.go
pallasathena92, line 386: A vLLM/sglang runtime that crashloops for hours can produce megabytes of logs per container instance. This function reads the entire log into memory on every reconcile. With 50 failing ISVCs requeuing every 30 seconds, you're doing 50 unbounded log fetches per reconcile cycle. This will spike controller memory and put unnecessary load on kubelet's log endpoint.

Fix:
```
tailLines := int64(200)
  limitBytes := int64(256 * 1024) // 256 KB
  req := sr.Clientset.CoreV1().Pods(namespace).GetLogs(podName, &v1.PodLogOptions{
      Container:  containerName,
      Previous:   previous,
      TailLines:  &tailLines,
      LimitBytes: &limitBytes,
  })
```
@slin1237 would you take another look for this?

op109lvb: You’re right. I was mostly thinking about the OOM + restart case, where `previous` logs get reset per container instance, so I underestimated the broader risk here. There are other crashloop/failure modes that can still produce larger logs, and this code was doing an unbounded fetch into memory on every reconcile. I updated the log fetch to cap both TailLines and LimitBytes so we avoid repeated unbounded reads while still keeping enough recent output for failure classification.

op109lvb: I also checked a sample OOM log, and the OOM signal was still present within the last 200 lines, so capping the fetch with TailLines looks like a reasonable safeguard here.

pkg/controller/v1beta1/inferenceservice/status/status_reconciler.go
op109lvb: Good catch. podLogFetcher was initialized even when clientset was nil, so tests that construct NewStatusReconciler(nil) could fall through into fetchPodLogs and emit spurious "clientset is not configured" errors. Since getContainerFailureMessage already skips log inspection when podLogFetcher is nil, it’s cleaner to only wire reconciler.fetchPodLogs when a real clientset is available.
Updated. Thanks.

pkg/controller/v1beta1/inferenceservice/components/router.go
op109lvb: I’d keep the current logger.WithName("InferenceServiceStatus") pattern. In logr, chained WithName calls preserve separate hierarchy segments, while "RouterReconciler.InferenceServiceStatus" is just one literal logger name. If simplification is needed, I'd use a statusLogger variable rather than flattening the name.

pkg/controller/v1beta1/inferenceservice/status/status_util.go
op109lvb: Good catch. I focused too narrowly on the OOM + restart case and missed that other crashloop failures can also produce large logs. I capped pod log fetches and validated by sample OOM logs. Thanks.

pallasathena92 (approved):

pkg/controller/v1beta1/inferenceservice/status/status_util.go
```diff
@@ -324,12 +331,103 @@ func (sr *StatusReconciler) checkContainerStatuses(status *v1beta1.InferenceServ
...
+
+	tailLines := sr.podLogTailLines
+	limitBytes := sr.podLogLimitBytes
+	req := sr.Clientset.CoreV1().Pods(namespace).GetLogs(podName, &v1.PodLogOptions{
+		Container:  containerName,
+		Previous:   previous,
+		TailLines:  &tailLines,
+		LimitBytes: &limitBytes,
+	})
+	logStream, err := req.Stream(context.Background())
```
YouNeedCryDear: maybe add a hard timeout for the log streaming?

YouNeedCryDear: Or even make this async?

op109lvb: Good catch. Added a 2-second timeout. Thanks

YouNeedCryDear (approved):

op109lvb:
New PR https://github.com/ome-projects/ome/pull/616, which uses pod status instead of fetching container logs.

PR #584, opened 2026-04-23
[OEP] Adding OEP for Multi-instance serving.

shenoyvvarun:
## What this PR does
-Support for Multi-instance GPU serving in OME. 
- NOTE: This OEP supports only hardware isolated Multi instance GPU serving. Other way is to support multiple inference service is via [KVCached](https://github.com/ovg-project/kvcached) which doesn't provide isolation but solves the case where a customer wants to serve multiple InferenceService via the same DAC.

## Why we need it
- Great opportunity to sell GPUs to customers who want predictable performance of DAC but, don't want to commit to full GPU.

Fixes #

## Checklist

### DIFF
--- a/oeps/007-multi-instance-serving/README.md
+++ b/oeps/007-multi-instance-serving/README.md
@@ -1,0 +1,639 @@
+# OEP-007: Multi-Instance Serving
+
+<!--
+This OEP describes how OME adds support for NVIDIA Multi-Instance GPU (MIG)
+serving using node-level MIG reconfiguration driven by InferenceService demand
+and applied through NVIDIA mig-parted.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Platform operator enables MIG-backed serving](#story-1-platform-operator-enables-mig-backed-serving)
+    - [Story 2: Application owner deploys a MIG-sized model](#story-2-application-owner-deploys-a-mig-sized-model)
+    - [Story 3: Accelerator-aware runtime selection chooses a MIG profile](#story-3-accelerator-aware-runtime-selection-chooses-a-mig-profile)
+    - [Story 4: OME cleans up MIG assignments safely](#story-4-ome-cleans-up-mig-assignments-safely)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Background](#background)
+  - [Architecture Overview](#architecture-overview)
+  - [Configuration Model](#configuration-model)
+  - [Controller Behavior](#controller-behavior)
+  - [MIG Manager Behavior](#mig-manager-behavior)
+  - [Resource and Scheduling Flow](#resource-and-scheduling-flow)
+  - [Operational Workflow](#operational-workflow)
+  - [API and Configuration Examples](#api-and-configuration-examples)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit Tests](#unit-tests)
+    - [Integration Tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This OEP introduces multi-instance serving in OME by adding support for NVIDIA
+Multi-Instance GPU (MIG) backed workloads. The design allows OME to recognize
+MIG resource requests in serving runtimes, map those requests to a desired MIG
+configuration, label an appropriate node with the requested configuration, and
+apply that configuration on the node through a dedicated `mig-manager`
+component that uses NVIDIA `mig-parted`.
+
+The feature targets an important gap in GPU utilization. Today, OME can deploy
+GPU-backed inference workloads, but it does not manage how a physical GPU is
+partitioned for smaller or more predictable serving units. With MIG, a single
+GPU can be split into multiple isolated GPU instances, each with dedicated
+compute and memory resources. That enables denser placement for smaller models,
+reduces over-allocation of full GPUs, and creates a more consistent serving
+shape for runtimes that are tuned to a specific GPU slice size. NVIDIA
+positions MIG as a hardware-based partitioning mechanism that provides strong
+isolation and predictable quality of service, while `mig-parted` provides a
+supported way to declaratively manage the desired MIG layout on a node.
+
+In OME, this capability is implemented as an operator-driven workflow rather
+than a new top-level API. Runtime authors request a specific MIG extended
+resource such as `nvidia.com/mig-2g.20gb`, the controller derives the
+corresponding MIG config name, selects an eligible node, persists the
+assignment in `InferenceService` annotations, and updates node labels so the
+node-local manager can converge the physical device state. The resulting design
+keeps the user-facing API small while fitting into Kubernetes-native scheduling
+and OME’s existing reconciliation model.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


YouNeedCryDear (commented):
High level question. How is it MIG going to fit in side the current AcceleratorClass? Are we going to have MIG defined as another set of accelerator?

oeps/007-multi-instance-serving/README.md
```diff
@@ -0,0 +1,639 @@
...
+enabled.
+
+The controller loop is responsible for discovering MIG demand from workload
+specifications, selecting a target node, and recording desired configuration.
+The node-local loop is responsible for turning that desired configuration into
+actual node device state.
+
+The main components are:
+
+- `pkg/controller/v1beta1/inferenceservice/mig_profile.go`: derives MIG demand,
```
YouNeedCryDear, line 304: Curious if we introduce mig_profile.go, does it means OME operator is Nvidia dependent?

oeps/007-multi-instance-serving/README.md
```diff
@@ -0,0 +1,639 @@
...
+Its responsibilities are:
+
+1. Watch the local node state and desired labels.
+2. Reconcile only when the node is marked as eligible for MIG dynamic
+   allocation.
+3. Load the allowed config names from the configured `mig-parted` file.
+4. Detect whether a node needs reconfiguration.
+5. Serialize changes with a local lock file so only one change runs at a time.
+6. Optionally inspect configured GPU clients before applying changes.
+7. Execute `nvidia-mig-parted apply -f <config> -c <desired>` with the host root
```
YouNeedCryDear, line 439: Is it possible we can re-use NVIDIA GPU Operator's native label-based partitioning? Instead of doing everything on our own? We probably should make it platform agnostic

oeps/007-multi-instance-serving/README.md
```diff
@@ -0,0 +1,639 @@
...
+positions MIG as a hardware-based partitioning mechanism that provides strong
+isolation and predictable quality of service, while `mig-parted` provides a
+supported way to declaratively manage the desired MIG layout on a node.
+
+In OME, this capability is implemented as an operator-driven workflow rather
+than a new top-level API. Runtime authors request a specific MIG extended
+resource such as `nvidia.com/mig-2g.20gb`, the controller derives the
+corresponding MIG config name, selects an eligible node, persists the
+assignment in `InferenceService` annotations, and updates node labels so the
+node-local manager can converge the physical device state. The resulting design
```
YouNeedCryDear, line 66: Once the inference service is gone do we want to persist the node labels or persist it?

oeps/007-multi-instance-serving/README.md
```diff
@@ -0,0 +1,639 @@
...
+  serialization, and optional reboot handling, but the operational experience is
+  still constrained by node state and NVIDIA driver requirements.
+- The current implementation uses a fixed mapping from requested resource name
+  to config name, such as `nvidia.com/mig-2g.20gb` -> `all-2g.20gb`. This keeps
+  the first iteration simple but does not yet model mixed-profile layouts.
+- Scheduling convergence depends on the device plugin refreshing allocatable
+  resources after MIG reconfiguration.
+- When dynamic allocation is disabled, OME does not perform node relabeling for
+  MIG reconfiguration and relies on preconfigured MIG nodes.
+- When dynamic allocation is enabled, OME selects only nodes explicitly marked
```
YouNeedCryDear, line 229: Is there a possibility that OME selects the node with MIG label but the node failed with other resource requirement such as CPU and mem?

oeps/007-multi-instance-serving/README.md
```diff
@@ -0,0 +1,639 @@
...
+   compatible nodes.
+3. Add a node-local MIG manager daemonset (provided out-of-box with OME) that
+   applies desired MIG configuration through `mig-parted`.
+4. Persist assignment state so OME can reconcile, update, and clean up MIG
+   configurations safely across `InferenceService` lifecycle events.
+5. Support an initial set of MIG profiles through a simple deterministic mapping
+   from requested resource name to node configuration name.
+6. Keep the design compatible with NVIDIA GPU Operator and device plugin based
+   clusters.
+7. Provide an option to disable OME-managed dynamic allocation entirely and use
```
YouNeedCryDear, line 111: What is the difference between the OME-managed Dynamic allocation and the NVIDIA DRA on k8s?

oeps/007-multi-instance-serving/README.md
shenoyvvarun: OME is working for mostly Nvidia accelartor but, has no lockin to a particular accelerator. I think for the first iteration we can rely on the GPU operator. But, we can make it such that its independent of gpu-operator.

oeps/007-multi-instance-serving/README.md
shenoyvvarun: Ack, for the first iteration we will simply use the nvidia-mig-manger. I will make it clear in the doc.

oeps/007-multi-instance-serving/README.md
shenoyvvarun: No its dynamically created and de-allocated

oeps/007-multi-instance-serving/README.md
shenoyvvarun: I mean the node selection logic needs to ensure all the resource requirements are met for a node.

oeps/007-multi-instance-serving/README.md
shenoyvvarun: There was no development on NVIDIA DRA for MIG for a year or so and last month allocation was supported as alpha. But, I believe they will not support A100 and older GPUs (I believe it makes business sense for them to support it newer GPUs). I am not sure whether this a acceptable drawback. But, rethinking, I am more leaning towards simply leveraging dra driver from nvidia.

shenoyvvarun:
> High level question. How is it MIG going to fit in side the current AcceleratorClass? Are we going to have MIG defined as another set of accelerator?

I think so, MIG devices would be another accelerator profile that would allow a runtime to customize the serving of a model. Do you have any concerns?

oeps/007-multi-instance-serving/README.md
YouNeedCryDear: If multiple MIG inference services are scheduled on the same node, the label should be there until the very last isvc is terminated correct?

YouNeedCryDear:
> > High level question. How is it MIG going to fit in side the current AcceleratorClass? Are we going to have MIG defined as another set of accelerator?
> 
> I think so, MIG devices would be another accelerator profile that would allow a runtime to customize the serving of a model. Do you have any concerns?

Only concern is there will be combination explosion 🤣 

Maybe we should do something on the accelerator class side as well? 

Also, MIG will be only available for models takes less than 1 GPU correct? Is it possible that a model will take 4 MIGs on 4 different GPU and call it TP4?

shenoyvvarun:
> > > High level question. How is it MIG going to fit in side the current AcceleratorClass? Are we going to have MIG defined as another set of accelerator?
> > 
> > 
> > I think so, MIG devices would be another accelerator profile that would allow a runtime to customize the serving of a model. Do you have any concerns?
> 
> Only concern is there will be combination explosion 🤣
> 
> Maybe we should do something on the accelerator class side as well?
> 
> Also, MIG will be only available for models takes less than 1 GPU correct? Is it possible that a model will take 4 MIGs on 4 different GPU and call it TP4?

No, that would defeat the purpose :) We should keep it simple MIG profile should always be TP=1, anything more just choose bigger MIG instance or GPU unit itself. Also If we support MIG via accelerator profile, it would be additive rather than combinatorial right?

oeps/007-multi-instance-serving/README.md
shenoyvvarun: Yeah, but the corresponding MIG labels will go away right? Consider this example, 
1 GPU is divided into:
- `nvidia.com/mig-2g.20gb` - LLama 3b
- `nvidia.com/mig-2g.20gb` - LLama 8b 

Now,customer with Llama 8b deletes the DAC.
- `nvidia.com/mig-2g.20gb` - LLama 3b
- `nvidia.com/mig-2g.20gb` - Label deleted and MIG device unallocated.

YouNeedCryDear (approved):

PR #595, opened 2026-04-30
[Core] add serving graph and predictor bridge

smfirmin:
## What this PR does

Adds a small shared serving-graph utility for controller reconciliation so the controller resolves the  active serving components from a single engine-first view.

  This PR:

  - resolves model, runtime, engine, decoder, and router through a shared serving-graph path
  - keeps predictor handling as a compatibility-only translation for predictor-only services during rollout
  - updates the controller to reconcile from the resolved serving graph instead of open-coding predictor-era  handling
  - adds focused tests for predictor compatibility translation, serving-graph resolution, entrypoint selection, and deployment-mode selection

  ## Why we need it

 The controller is moving toward an engine-first serving model where `engine` is the primary serving component, with optional `decoder` and optional `router`. Today, part of that logic is still split between direct controller wiring and predictor-era translation behavior.

This change is needed to make the controller resolve the active serving graph once and then use that shared result consistently for:

  - model/runtime resolution
  - predictor compatibility translation during rollout
  - merged component selection after runtime defaults are applied
  - deployment-mode and entrypoint selection for the resolved serving graph

  Without this, predictor compatibility and component-oriented reconciliation continue to be handled through
  separate paths, which increases controller complexity and makes follow-on work harder to reason about.

  ## How to test

  Run:

  ```bash
  env HOME=/tmp/oss-ome-home \
    KUBEBUILDER_ASSETS=/tmp/oss-ome-envtest-binaries/k8s/1.30.3-darwin-arm64 \
    GOCACHE=/tmp/oss-ome-gocache \
    GOMODCACHE=/tmp/oss-ome-gomodcache \
    GOPATH=/tmp/oss-ome-gopath \
    go test ./pkg/controller/v1beta1/inferenceservice/...

  What this covers:

  - predictor-only services are translated into the shared engine/model/runtime view for compatibility
  - serving-graph resolution identifies the effective engine/decoder/router set the controller should reconcile
  - router is selected as the entrypoint when present; otherwise engine remains the entrypoint
  - deployment modes are computed from the resolved graph after runtime defaults are merged
  - controller package tests continue to pass with the shared serving-graph path in place

  ## Checklist

  - [X] make test passes locally

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/controller.go
+++ b/pkg/controller/v1beta1/inferenceservice/controller.go
@@ -205,17 +205,27 @@ func (r *InferenceServiceReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// Determine which components to reconcile based on the spec
 	var reconcilers []components.Component
 
-	// Migrate predictor spec to new architecture if needed
-	if err := r.migratePredictorToNewArchitecture(isvc); err != nil {
-		r.Log.Error(err, "Failed to migrate predictor spec", "namespace", isvc.Namespace, "inferenceService", isvc.Name)
-		r.Recorder.Eventf(isvc, v1.EventTypeWarning, "PredictorMigrationError", err.Error())
+	servingGraph, err := isvcutils.ResolveServingGraph(isvc)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


YouNeedCryDear:
@pallasathena92 @beiguo218 Not sure if this is the right path going forward. I think we should fully deprecate predictor.

PR #596, opened 2026-04-30
[Core] support runtime-owned engine reconcile

smfirmin:
## What this PR does

 Enables the existing engine component reconciliation path to work when the engine is defined by the selected runtime rather than directly on the InferenceService.

  This keeps the change minimal and focused:
  - `MergeEngineSpec()` now preserves a runtime-owned engine spec when`isvc.spec.engine` is omitted.
  - Engine resource and affinity merging now treat a missing `isvc.spec.engine` as "no user override", so runtime defaults still flow into the reconciled engine workload.
  - External service selection now recognizes runtime-driven engine deployments and routes traffic to the engine component when appropriate.
  - Adds focused regression coverage for runtime-owned engine reconciliation, external service targeting, and engine resource/affinity merging.

  ## Why we need it

Some serving runtimes define their engine configuration in`runtime.spec.engineConfig` without requiring users to also duplicate that  configuration in `InferenceService.spec.engine`.

Before this change, those runtime-owned engine cases could miss parts of the normal engine reconciliation flow because several code paths assumed the engine only existed when `isvc.spec.engine` was set directly. That made  runtime-driven engine deployments incomplete, especially for multi-node engine reconciliation and external service targeting.

  ## How to test

  1. Run the full repo test suite:
     ```bash
     make test

  2. Focus on the added regression coverage under:

     go test ./pkg/controller/v1beta1/inferenceservice/...
  3. Verify that an InferenceService using a runtime with
     runtime.spec.engineConfig and no explicit isvc.spec.engine still:
      - reconciles an engine component
      - applies runtime/default engine resources and affinity when the user did
        not override them
      - targets the engine component for external service selection when no
        router is present

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/components/base.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/base.go
@@ -321,9 +321,9 @@ func MergeResources(b *BaseComponentFields, container *corev1.Container) {
 // resources are respected and not overridden, while providing sensible defaults from the
 // runtime and accelerator class when resources are not specified.
 func MergeEngineResources(b *BaseComponentFields, isvc *v1beta1.InferenceService, container *corev1.Container) {
-	if isvc.Spec.Engine != nil &&
-		(isvc.Spec.Engine.Runner == nil ||
-			isResourcesUnspecified(isvc.Spec.Engine.Runner.Container.Resources)) {
+	if isvc.Spec.Engine == nil ||
+		isvc.Spec.Engine.Runner == nil ||
+		isResourcesUnspecified(isvc.Spec.Engine.Runner.Container.Resources) {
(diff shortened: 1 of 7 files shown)


YouNeedCryDear:
@smfirmin I don't think this is the expected behavior. The component specs are defined in the runtime but the isvc is mandatory to define the replicas or optionally overwrite the spec. For example, when router spec is defined in the runtime, but no router in isvc, then isvc shouldn't spin up a router component.

smfirmin:
@YouNeedCryDear Thank you for the review - I agree, this PR is crossing a contract boundary.

The actual goal was to make the generic multi-node engine behavior reconcile through the shared engine component path, so we stop depending on predictor-first reconciliation for that behavior. It was not to change component activation semantics so that runtime.spec.engineConfig alone causes an engine component to be created.

I'll close this PR

PR #602, opened 2026-05-08
fix(modelconfig): return empty architecture when Mistral Architectures is missing

### LINKED ISSUE #601: bug: MistralForCausalLM models misclassified as EMBEDDING due to GetArchitecture() fallback
## Summary

When importing `mistralai/Mistral-7B-Instruct-v0.3` or `mistralai/Mistral-Nemo-Instruct-2407` from HuggingFace, OME classifies both as `EMBEDDING` instead of `TEXT_GENERATION`. Chat/completions requests against the resulting endpoint return:

```
400 Bad Request: importedModel does not support any of: [TextToText, ImageTextToText]
```

Both models have `"architectures": ["MistralForCausalLM"]` in their HF `config.json` and should be `TEXT_GENERATION`.

## Root Cause

Two code locations interact to produce the bug:

**1. `pkg/hfutil/modelconfig/mistral.go` — `GetArchitecture()` fallback**

```go
func (c *MistralConfig) GetArchitecture() string {
    if len(c.Architectures) > 0 {
        return c.Architectures[0]
    }
    return "MistralModel"   // ← dangerous fallback
}
```

If `Architectures` is empty (e.g. JSON parsing fails, field missing, or struct mismatch), the method silently returns `"MistralModel"`.

**2. `pkg/modelagent/config_parser.go` — `determineModelCapabilitiesFromHF()`**

```go
if strings.Contains(strings.ToLower(architecture), "embedding") ||
    strings.Contains(strings.ToLower(architecture), "sentence") ||
    strings.Contains(strings.ToLower(modelType), "bert") ||
    // Special case for known embedding models
    (strings.Contains(strings.ToLower(modelType), "mistral") &&
        strings.Contains(strings.ToLower(architecture), "mistralmodel")) {
    return append(capabilities, string(v1beta1.ModelCapabilityEmbedding))
}
```

When the fallback fires, `modelType = "mistral"` and `architecture = "MistralModel"` satisfy the special-case condition, and the model is classified as `EMBEDDING`.

The intended path for `intfloat/e5-mistral-7b-instruct` (a genuine embedding model) is correct: its HF config has `architectures: []` or uses the base `MistralModel` architecture, so the fallback correctly labels it. The problem is that causal-LM models whose `Architectures` field fails to populate get the same treatment.

## Repro

Import either of these models via the OME model-agent and check the resulting `ClusterBaseModel.spec.modelCapabilities`:

- `mistralai/Mistral-7B-Instruct-v0.3` — `architectures: ["MistralForCausalLM"]` — classified as `EMBEDDING` ❌
- `mistralai/Mistral-Nemo-Instruct-2407` — `architectures: ["MistralForCausalLM"]` — classified as `EMBEDDING` ❌
- `intfloat/e5-mistral-7b-instruct` — embedding model — classified as `EMBEDDING` ✅

## Expected Behaviour

| Model | Architecture (HF) | Expected capability |
|---|---|---|
| `mistralai/Mistral-7B-Instruct-v0.3` | `MistralForCausalLM` | `TEXT_GENERATION` |
| `mistralai/Mistral-Nemo-Instruct-2407` | `MistralForCausalLM` | `TEXT_GENERATION` |
| `intfloat/e5-mistral-7b-instruct` | `MistralModel` | `EMBEDDING` |

## Proposed Fix

Change the `GetArchitecture()` fallback from `"MistralModel"` to `""` so a missing/unparsed `Architectures` field does not accidentally satisfy the embedding special-case:

```go
func (c *MistralConfig) GetArchitecture() string {
    if len(c.Architectures) > 0 {
        return c.Architectures[0]
    }
    return ""   // don't assume MistralModel; let caller treat as unknown
}
```

Alternatively, tighten the special-case check in `config_parser.go` to require the architecture to be exactly `"MistralModel"` (case-insensitive) rather than a substring match, and only when `Architectures` was explicitly set (not via fallback).

## Additional Context

- `autoSelect` is `false` on the `vllm-e5-mistral-7b-instruct` runtime and the two runtimes use distinct `modelArchitecture` values (`MistralModel` vs `MistralForCausalLM`), so **runtime auto-selection is not affected** — the runtimes cannot be confused with each other.
- The misclassification only affects capability gating at the endpoint level (chat vs embedding API routing).

abinggo:
## Summary

Fixes #601 — MistralForCausalLM models misclassified as EMBEDDING due to `GetArchitecture()` fallback.

## Root Cause

When a Mistral model's `config.json` lacks the `architectures` field:
1. `GetArchitecture()` returned `"MistralModel"` (dangerous fallback)
2. `config_parser.go`'s embedding detection matched `(modelType contains "mistral") && (architecture contains "mistralmodel")`
3. `IsEmbedding()` returned `true` unconditionally

Result: standard Mistral causal LM models (e.g. `mistralai/Mistral-7B-Instruct-v0.3`) were classified as `EMBEDDING`.

## Fix

- **`GetArchitecture()`**: Return `""` (empty string) when `Architectures` is empty — consistent with `BaseModelConfig` and other model configs (phi, deepseek_v3)
- **`IsEmbedding()`**: Return `false` — MistralConfig is used for all Mistral-based models, not just embedding ones

## Why genuine embedding models still work

Models like `intfloat/e5-mistral-7b-instruct` explicitly set `"architectures": ["MistralModel"]` in their `config.json`. With this fix:
- Their `GetArchitecture()` returns `"MistralModel"` from the explicit field (not a fallback)
- The config_parser special case `(mistral + mistralmodel)` still matches correctly

## Audit of other models

Checked all `GetArchitecture()` implementations. Other models with non-empty fallbacks (`mixtral` → `"MixtralForCausalLM"`, `llama` → `"LlamaForCausalLM"`, `dbrx` → `"DbrxForCausalLM"`) use `*ForCausalLM` suffixes that don't trigger embedding detection. Only `mistral`'s `"MistralModel"` was dangerous.

## Test plan

- [x] Added `TestMistralConfigNoArchitectures` — verifies `GetArchitecture()` returns `""` and `IsEmbedding()` returns `false` when Architectures is missing
- [x] Added `testdata/mistral_no_architectures.json` fixture
- [x] Existing tests pass (standard Mistral-7B and Mistral-7B-Instruct configs)

### DIFF
--- a/pkg/hfutil/modelconfig/mistral.go
+++ b/pkg/hfutil/modelconfig/mistral.go
@@ -117,7 +117,7 @@ func (c *MistralConfig) GetArchitecture() string {
 	if len(c.Architectures) > 0 {
 		return c.Architectures[0]
 	}
-	return "MistralModel"
+	return ""
 }
 
 func (c *MistralConfig) GetModelType() string {
@@ -140,9 +140,12 @@ func (c *MistralConfig) HasVision() bool {
 	return false
 }
 
-// IsEmbedding returns true since this is an embedding model
+// IsEmbedding returns true only when the architecture explicitly indicates an embedding model.
+// Genuine embedding models like intfloat/e5-mistral-7b-instruct set Architectures to
+// ["MistralModel"] in their config.json, which the config_parser recognizes via the
+// (modelType=="mistral" && architecture=="mistralmodel") special case.
 func (c *MistralConfig) IsEmbedding() bool {
-	return true
+	return false
 }
 
 // Register the Mistral model handler
--- a/pkg/hfutil/modelconfig/mistral.go
+++ b/pkg/hfutil/modelconfig/mistral.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 )
 
 // MistralConfig defines the configuration for Mistral models
@@ -140,12 +141,10 @@ func (c *MistralConfig) HasVision() bool {
 	return false
 }
 
-// IsEmbedding returns true only when the architecture explicitly indicates an embedding model.
-// Genuine embedding models like intfloat/e5-mistral-7b-instruct set Architectures to
-// ["MistralModel"] in their config.json, which the config_parser recognizes via the
-// (modelType=="mistral" && architecture=="mistralmodel") special case.
+// IsEmbedding returns true when the architecture is "MistralModel", which indicates
+// an embedding model (e.g. intfloat/e5-mistral-7b-instruct).
 func (c *MistralConfig) IsEmbedding() bool {
-	return false
+	return strings.EqualFold(c.GetArchitecture(), "MistralModel")
 }
 
 // Register the Mistral model handler
--- a/pkg/hfutil/modelconfig/mistral_test.go
+++ b/pkg/hfutil/modelconfig/mistral_test.go
@@ -84,6 +84,36 @@ func TestLoadModelWithMistral(t *testing.T) {
 	t.Logf("Mistral model parameter count via generic loader: %s", FormatParamCount(paramCount))
 }
 
+// TestMistralConfigNoArchitectures verifies that when the Architectures field
+// is missing from config.json, GetArchitecture() returns an empty string rather
+// than a misleading fallback like "MistralModel" which would cause the model to
+// be misclassified as an embedding model by the config parser.
+// Regression test for https://github.com/ome-projects/ome/issues/601
+func TestMistralConfigNoArchitectures(t *testing.T) {
+	configPath := filepath.Join("testdata", "mistral_no_architectures.json")
+
+	config, err := LoadModelConfig(configPath)
+	if err != nil {
+		t.Fatalf("Failed to load Mistral config without architectures: %v", err)
+	}
+
+	// GetArchitecture must return empty string when Architectures is absent
+	arch := config.GetArchitecture()
+	if arch != "" {
+		t.Errorf("Expected empty architecture when Architectures field is missing, got '%s'", arch)
+	}
+
+	// IsEmbedding must return false for a generic Mistral model
+	if config.IsEmbedding() {
+		t.Error("Expected IsEmbedding() to return false for a generic Mistral model without explicit embedding architecture")
+	}
+
+	// Verify it's still recognized as a Mistral model
+	if config.GetModelType() != "mistral" {
+		t.Errorf("Expected model type 'mistral', got '%s'", config.GetModelType())
+	}
+}
+
 func TestMistralInstructConfig(t *testing.T) {
 	configPath := filepath.Join("testdata", "mistral_7b_instruct.json")
 
--- a/pkg/hfutil/modelconfig/mistral_test.go
+++ b/pkg/hfutil/modelconfig/mistral_test.go
@@ -114,6 +114,35 @@ func TestMistralConfigNoArchitectures(t *testing.T) {
 	}
 }
 
+// TestMistralConfigEmbedding verifies that a Mistral model with
+// "architectures": ["MistralModel"] is correctly identified as an embedding
+// model (e.g. intfloat/e5-mistral-7b-instruct).
+// This is the load-bearing positive case for config_parser.go's EMBEDDING classification.
+func TestMistralConfigEmbedding(t *testing.T) {
+	configPath := filepath.Join("testdata", "mistral_embedding.json")
+
+	config, err := LoadModelConfig(configPath)
+	if err != nil {
+		t.Fatalf("Failed to load Mistral embedding config: %v", err)
+	}
+
+	// GetArchitecture must return "MistralModel"
+	arch := config.GetArchitecture()
+	if arch != "MistralModel" {
+		t.Errorf("Expected architecture 'MistralModel', got '%s'", arch)
+	}
+
+	// GetModelType must return "mistral"
+	if config.GetModelType() != "mistral" {
+		t.Errorf("Expected model type 'mistral', got '%s'", config.GetModelType())
+	}
+
+	// IsEmbedding must return true for MistralModel architecture
+	if !config.IsEmbedding() {
+		t.Error("Expected IsEmbedding() to return true for a model with 'MistralModel' architecture")
+	}
+}
+
 func TestMistralInstructConfig(t *testing.T) {
 	configPath := filepath.Join("testdata", "mistral_7b_instruct.json")
 
--- a/pkg/hfutil/modelconfig/testdata/mistral_embedding.json
+++ b/pkg/hfutil/modelconfig/testdata/mistral_embedding.json
@@ -1,0 +1,22 @@
+{
+  "architectures": ["MistralModel"],
+  "bos_token_id": 1,
+  "eos_token_id": 2,
+  "hidden_act": "silu",
+  "hidden_size": 4096,
+  "initializer_range": 0.02,
+  "intermediate_size": 14336,
+  "max_position_embeddings": 32768,
+  "model_type": "mistral",
+  "num_attention_heads": 32,
+  "num_hidden_layers": 32,
+  "num_key_value_heads": 8,
+  "rms_norm_eps": 1e-05,
+  "rope_theta": 10000.0,
+  "sliding_window": 4096,
+  "tie_word_embeddings": false,
+  "torch_dtype": "float16",
+  "transformers_version": "4.36.0",
+  "use_cache": true,
+  "vocab_size": 32000
+}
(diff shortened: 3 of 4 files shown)


YouNeedCryDear:
@OpheliaLjh would you mind quickly overlook this PR?

pkg/hfutil/modelconfig/mistral.go
```diff
@@ -140,9 +140,12 @@ func (c *MistralConfig) HasVision() bool {
...
 }
 
-// IsEmbedding returns true since this is an embedding model
+// IsEmbedding returns true only when the architecture explicitly indicates an embedding model.
+// Genuine embedding models like intfloat/e5-mistral-7b-instruct set Architectures to
+// ["MistralModel"] in their config.json, which the config_parser recognizes via the
+// (modelType=="mistral" && architecture=="mistralmodel") special case.
 func (c *MistralConfig) IsEmbedding() bool {
-	return true
+	return false
```
OpheliaLjh: The comment here is misleading. It says `IsEmbedding` _“returns true only when the architecture explicitly indicates an embedding model”_ but the implementation unconditionally returns `false`. The architecture-based check the comment is describing lives in `config_parser.go`'s `(mistral && mistralmodel)` condition, not here.

Two clean options:

**Option A** — implement what the comment describes, which also lets you remove the special case from `config_parser.go` entirely:
```go
func (c *MistralConfig) IsEmbedding() bool {
    return c.GetArchitecture() == "MistralModel"
}
```

**Option B** — keep `return false` but rewrite the comment to say so plainly:
```go
// IsEmbedding returns false. MistralConfig covers all Mistral-derived architectures,
// not just embedding models. Genuine embedding models (e.g. e5-mistral-7b-instruct)
// are detected via the (mistral && mistralmodel) architecture match in config_parser.go.
func (c *MistralConfig) IsEmbedding() bool {
    return false
}
```

pkg/hfutil/modelconfig/mistral_test.go
```diff
@@ -84,6 +84,36 @@ func TestLoadModelWithMistral(t *testing.T) {
...
+	// IsEmbedding must return false for a generic Mistral model
+	if config.IsEmbedding() {
+		t.Error("Expected IsEmbedding() to return false for a generic Mistral model without explicit embedding architecture")
+	}
+
+	// Verify it's still recognized as a Mistral model
+	if config.GetModelType() != "mistral" {
+		t.Errorf("Expected model type 'mistral', got '%s'", config.GetModelType())
+	}
+}
```
OpheliaLjh, line 115: This test covers the negative case (no architectures → not embedding) but there is no counterpart for the positive case. Please add a test with a fixture that has `"architectures": ["MistralModel"]` set and assert:
- `GetArchitecture() == "MistralModel"`
- `GetModelType() == "mistral"`

That combination is what `config_parser.go` relies on to classify e5-mistral as EMBEDDING — it is the load-bearing path and easy to accidentally break in a future refactor. Without a test here, nothing catches a regression.

OpheliaLjh (commented):
Fix is conceptually correct — the root-cause behaviour after the change is right for both affected models:
- `mistral-7b-instruct` (no `architectures` field in `config.json`): `GetArchitecture()` now returns `""`, the `(mistral && mistralmodel)` condition in `determineModelCapabilitiesFromHF` evaluates false → TEXT\_TO\_TEXT ✓
- `e5-mistral-7b-instruct` (has `"architectures": ["MistralModel"]` explicitly): `GetArchitecture()` still returns `"MistralModel"`, that condition fires → EMBEDDING ✓

Two things worth addressing before merging:

**1. `IsEmbedding()` comment contradicts the implementation.** The new docstring says _“returns true only when the architecture explicitly indicates an embedding model”_ but the body unconditionally returns `false`. `config_parser.go` checks `hfModel.IsEmbedding()` first in the OR chain — a reader looking only at this method would conclude e5-mistral flows through `IsEmbedding()`. It does not; it falls through to the `(mistral && mistralmodel)` string match. See inline.

**2. Missing regression test for the embedding-positive case.** The new test covers the no-architecture path (mistral-7b scenario) but there is nothing asserting that a model with `"architectures": ["MistralModel"]` set still resolves to `EMBEDDING` via `config_parser`. That is exactly the path e5-mistral depends on and the one most at risk of being quietly broken by a future cleanup. See inline on the test file.

pkg/hfutil/modelconfig/mistral.go
OpheliaLjh: The comment here is misleading. It says `IsEmbedding` _"returns true only when the architecture explicitly indicates an embedding model"_ but the implementation unconditionally returns `false`. The architecture-based check the comment is describing lives in `config_parser.go`'s `(mistral && mistralmodel)` condition, not here.

Two clean options:

**Option A** — implement what the comment describes, which also lets you remove the special case from `config_parser.go` entirely:
```go
func (c *MistralConfig) IsEmbedding() bool {
	return c.GetArchitecture() == "MistralModel"
}
```

**Option B** — keep `return false` but rewrite the comment to say so plainly:
```go
// IsEmbedding returns false. MistralConfig covers all Mistral-derived architectures,
// not just embedding models. Genuine embedding models (e.g. e5-mistral-7b-instruct)
// are detected via the (mistral && mistralmodel) architecture match in config_parser.go.
func (c *MistralConfig) IsEmbedding() bool {
	return false
}
```

pkg/hfutil/modelconfig/mistral_test.go
OpheliaLjh, line 115: This test covers the negative case (no architectures → not embedding) but there is no counterpart for the positive case. Please add a test with a fixture that has `"architectures": ["MistralModel"]` set and assert:
- `GetArchitecture() == "MistralModel"`
- `GetModelType() == "mistral"`

That combination is what `config_parser.go` relies on to classify e5-mistral as EMBEDDING — it is the load-bearing path and easy to accidentally break in a future refactor. Without a test here, nothing catches a regression.

abinggo:
Thanks for the detailed review! Addressed both points:

1. **IsEmbedding()** — went with Option A: `return strings.EqualFold(c.GetArchitecture(), "MistralModel")`. The logic is now self-contained in MistralConfig.

2. **Embedding-positive test** — added `TestMistralConfigEmbedding` with a `mistral_embedding.json` fixture that asserts `GetArchitecture() == "MistralModel"`, `GetModelType() == "mistral"`, and `IsEmbedding() == true`.

pkg/hfutil/modelconfig/mistral.go
```diff
@@ -117,7 +118,7 @@ func (c *MistralConfig) GetArchitecture() string {
 	if len(c.Architectures) > 0 {
 		return c.Architectures[0]
 	}
-	return "MistralModel"
+	return ""
```
YouNeedCryDear, line 121: How is the empty string as architecture going to be handled by the downstream matcher? Does the corresponding runtime have explicit "" as architecture?

abinggo: @YouNeedCryDear  Thanks for the review. I verified the downstream handling in pkg/modelagent/config_parser.go — all usages of GetArchitecture() normalize via strings.ToLower() and use strings.Contains() for pattern matching. When architecture is "", no specific patterns match, and determineModelCapabilitiesFromHF() correctly falls through to the default ModelCapabilityTextToText at line 586. The embedding check at line 575 relies on IsEmbedding() (which is now self-contained in MistralConfig) rather than the architecture string directly, so this is safe.
This is also consistent with how BaseModelConfig, PhiConfig, and DeepSeekV3Config handle missing architectures — they all return "".

YouNeedCryDear (approved):

OpheliaLjh (approved):
LGTM. This addresses the main functionality concern: Mistral configs without
`architectures` no longer fall back to `MistralModel`, while explicit `MistralModel`
embedding configs still classify correctly. The added no-architecture and positive embedding tests cover the important paths.

abinggo:
Hi @slin1237 — this PR has approvals from @YouNeedCryDear and @OpheliaLjh, and the branch is mergeable but appears to be blocked on a CODEOWNERS sign-off. Whenever you have a moment, could you take a quick look? Happy to address anything that comes up. Thanks!

abinggo:
Hi @slin1237 — friendly bump on this. PR has approvals from @YouNeedCryDear and @OpheliaLjh, branch is mergeable, all checks passing. Whenever you have a moment for the codeowner sign-off, would appreciate it. Thanks!

abinggo:
Hi @slin1237 @pallasathena92 @beiguo218 — friendly bump on this PR.

Status:
- approved by @OpheliaLjh and @YouNeedCryDear (2026-05-12)
- branch is mergeable
- all CI checks pass
- `mergeStateStatus` is `BLOCKED` waiting for `pkg/hfutil/` codeowner review

The change is small (+107/-3, scoped to `mistral.go` fallback handling for missing `architectures` field) and has been sitting since May 8. Would appreciate a quick look from any of the codeowners — happy to address any concerns.

PR #611, opened 2026-05-13
[Fix] Override data parallel size for accelerator configs

YouNeedCryDear:
## What this PR does

Adds accelerator-specific runtime argument overrides for data parallel size, covering `--dp-size`, `--dp`, `--data-parallel-size`, and `-dp`. It also extends tensor and pipeline parallel overrides to handle `-tp` and `-pp` aliases.

## Why we need it

`tensorParallelismOverride.dataParallelSize` is part of the ServingRuntime accelerator config API, but the runtime argument override logic only applied tensor and pipeline parallel sizes. As a result, accelerator configs could set `dataParallelSize` without changing the generated runtime command. This makes DP override behavior match TP and PP override behavior.

Fixes # (not filed)

## How to test

```bash
env GOCACHE=/private/tmp/ome-go-build-cache go test ./pkg/controller/v1beta1/inferenceservice/components -count=1
```

## Checklist

- [x] Docs updated (if applicable) - N/A

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/components/base.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/base.go
@@ -270,16 +270,22 @@ func MergeRuntimeArgumentsOverride(b *BaseComponentFields, container *corev1.Con
 				tensorParallelismConfig := acceleratorModelConfig.TensorParallelismOverride
 
 				// Override tensor parallel size if specified
-				// --tp-size and --tp are parameters used in sglang
-				// --tensor-parallel-size is the parameter used in vllm
+				// --tp-size and --tp are used in sglang
+				// --tensor-parallel-size and -tp are used in vllm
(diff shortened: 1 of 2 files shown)


pallasathena92 (approved):

beiguo218 (approved):

YouNeedCryDear:
@slin1237 Let me know when this can be merged.

heymrbox (commented):
Reviewed the TP/PP/DP override patch. I did not find a blocking issue.

Verification run on the PR head:

```sh
go test ./pkg/controller/v1beta1/inferenceservice/components -run TestMergeRuntimeArgumentsOverride -count=1
```

The DP override follows the existing `overrideParam` behavior used for TP/PP, and the added tests cover long flags, SGLang size flags, command-based shorthand flags, and single-dash aliases. PR checks are green as of this review.

PR #616, opened 2026-05-19
[Core] Expose container startup failures in InferenceService status

op109lvb:
## What this PR does

Adds `ContainerStartupFailed` to InferenceService status for serving containers that are unready in `CrashLoopBackOff` past the restart threshold.

### More details 
1. Why this approach?
This PR uses pod status instead of fetching container logs during reconciliation. `CrashLoopBackOff` is a bounded, actionable signal for the most common serving-container startup failure pattern. Detailed debugging stays in pod events, pod logs, and controller logs.

2. Behavior

- `ModelLoadFailed` remains for model download, model load, and init-container failures.
- `ContainerStartupFailed` applies only to Engine/Decoder serving containers backed by a `BaseModel` or `ClusterBaseModel`.
- No new condition is added; model state remains `FailedToLoad` and transition status remains `BlockedByFailedLoad`.
- Raw deployment handling stays conservative after a model has already loaded, because it cannot reliably prove whether a crash happened before readiness or after serving.

## Why we need it

When a serving container keeps restarting, we need to distinguish startup failures from runtime failures.

If the container never becomes ready, the rollout is blocked and higher-level automation can remove the failed deployment to avoid wasting resources.

This PR makes startup failures visible in InferenceService status instead of letting them remain pending or appear as generic model load failures.

## How to test
Check isvc status when container cannot start due to OOM. 

Sample container status: 
```
status:
initContainerStatuses:
- name: ome-container
      state:
        waiting:
          reason: CrashLoopBackOff
          message: >-
            back-off 5m0s restarting failed container=ome-container
```
Sample isvc status: 
```
status:
modelStatus:
    lastFailureInfo:
      location: >-
        <model-name>/ome-container
      modelRevisionName: '1'
      reason: ContainerStartupFailed
    modelRevisionStates:
      activeModelState: ''
      targetModelState: FailedToLoad
    transitionStatus: BlockedByFailedLoad
```

## Checklist

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/components/base.go
+++ b/pkg/controller/v1beta1/inferenceservice/components/base.go
@@ -494,7 +494,26 @@ func UpdateComponentStatus(b *BaseComponentFields, isvc *v1beta1.InferenceServic
 	if err != nil {
 		return errors.Wrapf(err, "failed to list %s pods by label", componentType)
 	}
-	b.StatusManager.PropagateModelStatus(&isvc.Status, statusSpec, pods, rawDeployment)
+	reportContainerStartupFailure := shouldReportContainerStartupFailure(b, isvc, componentType)
+	b.StatusManager.PropagateModelStatus(&isvc.Status, statusSpec, pods, rawDeployment, reportContainerStartupFailure)
 
 	return nil
+}
+
+func shouldReportContainerStartupFailure(b *BaseComponentFields, isvc *v1beta1.InferenceService, componentType v1beta1.ComponentType) bool {
+	if componentType != v1beta1.EngineComponent && componentType != v1beta1.DecoderComponent {
+		return false
+	}
+	if b == nil || b.BaseModelMeta == nil || isvc == nil || isvc.Spec.Model == nil {
+		return false
+	}
+	if isvc.Spec.Model.Kind != nil &&
+		*isvc.Spec.Model.Kind != constants.BaseModel &&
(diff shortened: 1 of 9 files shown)


pkg/controller/v1beta1/inferenceservice/components/base.go
```diff
@@ -494,7 +494,26 @@ func UpdateComponentStatus(b *BaseComponentFields, isvc *v1beta1.InferenceServic
...
 	}
-	b.StatusManager.PropagateModelStatus(&isvc.Status, statusSpec, pods, rawDeployment)
+	reportContainerStartupFailure := shouldReportContainerStartupFailure(b, isvc, componentType)
+	b.StatusManager.PropagateModelStatus(&isvc.Status, statusSpec, pods, rawDeployment, reportContainerStartupFailure)
 
 	return nil
 }
+
+func shouldReportContainerStartupFailure(b *BaseComponentFields, isvc *v1beta1.InferenceService, componentType v1beta1.ComponentType) bool {
+	if componentType != v1beta1.EngineComponent && componentType != v1beta1.DecoderComponent {
```
YouNeedCryDear, line 504: how about router?

YouNeedCryDear (approved):

pkg/controller/v1beta1/inferenceservice/components/base.go
op109lvb: Router is not a model serving component and it does not load BaseModel.

slin1237 (approved):

PR #619, opened 2026-05-20
[Bugfix] Return pod webhook errors instead of panics

heymrbox:
## What this PR does

Handle malformed pod injector config and invalid resource quantities as
admission errors instead of panicking in the webhook process.

Also return an explicit error when TensorRT model init cannot determine the
main container GPU limit, and add regression tests for those failure paths.
## Why we need it

Fixes pod webhook panic issue. The error should be properly handled.

## How to test

make test

## Checklist

### DIFF
--- a/pkg/webhook/admission/pod/fine_tuned_adapter_injector.go
+++ b/pkg/webhook/admission/pod/fine_tuned_adapter_injector.go
@@ -71,7 +71,10 @@ func (fa *FineTunedAdapterInjector) injectFineTunedAdapter(pod *v1.Pod) error {
 
 	modelInitMounts := fa.getVolumeMounts(pod)
 
-	fineTunedWeightUri, _ := fa.getFineTunedWeightUri(pod)
+	fineTunedWeightUri, err := fa.getFineTunedWeightUri(pod)
+	if err != nil {
+		return err
+	}
 
 	initEnvs, err := fa.getModelInitEnvs(pod, fineTunedWeightUri)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/webhook/admission/pod/fine_tuned_adapter_injector.go
YouNeedCryDear: This error still got discarded and it might cause panic during the env build when it is nil

YouNeedCryDear, line 148: Do we have a specific check to ensure this storageUri exist?

heymrbox: Fixed in ddfa9bb: getFineTunedWeightUri now explicitly rejects missing or empty spec.storage.storageUri before parsing. Added regression coverage for missing storage, missing storageUri, and empty storageUri.

heymrbox: Fixed in ddfa9bb: injectFineTunedAdapter now returns the getFineTunedWeightUri error instead of discarding it, so a missing FineTunedWeight is surfaced as an admission error. Added TestFineTunedAdapterReturnsWeightLookupError for that path.

pkg/webhook/admission/pod/serving_sidecar_injector.go
```diff
@@ -159,26 +161,22 @@ func (ss *ServingSidecarInjector) getServingSidecarEnvs(fineTunedWeightFTStrateg
 }
 
 // createServingSidecarContainer constructs the serving sidecar configuration.
-func (ss *ServingSidecarInjector) createServingSidecarContainer(envs []v1.EnvVar, mounts []v1.VolumeMount, securityContext *v1.SecurityContext) *v1.Container {
+func (ss *ServingSidecarInjector) createServingSidecarContainer(envs []v1.EnvVar, mounts []v1.VolumeMount, securityContext *v1.SecurityContext) (*v1.Container, error) {
+	resources, err := newResourceRequirements(ss.CpuLimit, ss.MemoryLimit, ss.CpuRequest, ss.MemoryRequest)
```
pallasathena92, line 165: Can you clarify the benefit of this change? It looks like we are replacing a small 12-line block with 50+ lines, so I want to make sure the added complexity is justified.

heymrbox: Clarification: the helper is intended to remove the resource.MustParse panic path across all three pod injectors, not just this sidecar block. These values come from ConfigMap JSON, so malformed or missing quantities would previously panic the admission webhook process; newResourceRequirements lets model-init, fine-tuned adapter, and serving-sidecar all return a field-specific admission error instead. I kept it shared so the three injectors use the same validation behavior rather than carrying three duplicated ParseQuantity blocks.

pallasathena92: OK， i found it. modei_init_injector reuse this function as well.

pallasathena92 (approved):

PR #623, opened 2026-05-26
[Core] Add data parallelism override support

smfirmin:
## What this PR does

Adds explicit data parallelism override support for accelerator-specific runtime configuration.

When `tensorParallelismOverride.dataParallelSize` is set for a selected accelerator class, OME now updates matching runtime arguments using the supported SGLang and vLLM aliases:

- `--dp-size`
- `--dp`
- `--data-parallel-size`

This mirrors the existing tensor and pipeline parallelism override behavior.

## Why we need it

The API already exposes `dataParallelSize`, but the reconciler only applied tensor parallel and pipeline parallel sizing. This meant accelerator-specific data parallel sizing could be configured but would not be reflected in generated runtime command or args.

Data parallel sizing is needed for MoE deployments where effective parallelism depends on TP x DP.

Fixes #

## How to test

```bash
go test ./pkg/controller/v1beta1/inferenceservice/components
git diff --check
```

## Checklist

- [X] Tests added/updated (if applicable)
- [X] Docs updated (if applicable)


YouNeedCryDear:
duplicate of https://github.com/ome-projects/ome/pull/611

PR #629, opened 2026-06-12
Skip OCI Object Storage download for Ready same-path artifacts

op109lvb:
## What this PR does

Speeds up copied model reconciliation when the copied CR uses OCI Object Storage and points to the same storage path as a model artifact that is already Ready on the same node.

For same-path object storage reuse, model-agent skips the expensive download/integrity-validation path instead of re-running file validation.

The implementation looks at this node’s Ready model entries and reuses only when the storage URI, path, schema path, storage key, parameters, and resolved destination path match exactly, and the local destination path already exists.

## Why we need it

Copied model CRs can point to the same object storage path as an already Ready model on the same node. In that case, re-downloading or re-validating all files adds unnecessary reconciliation latency.

This speeds up copied model CR reconciliation while preserving safety:

- reuse only happens for OCI Object Storage models
- reuse only happens when the storage URI/path/params match exactly
- reuse only happens if the local destination path exists on that node
- nodes without a local Ready artifact continue through the normal download/validation flow
- Hugging Face continues using its existing SHA-based artifact reuse logic

## How to test
```
go test -count=1 ./pkg/modelagent ./pkg/controller/v1beta1/basemodel
ok  	sigs.k8s.io/ome/pkg/modelagent	1.715s
ok  	sigs.k8s.io/ome/pkg/controller/v1beta1/basemodel	0.493s
```
## Checklist

### DIFF
--- a/pkg/modelagent/gopher.go
+++ b/pkg/modelagent/gopher.go
@@ -2,6 +2,7 @@ package modelagent
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -324,31 +325,35 @@ func (s *Gopher) processTask(task *GopherTask) error {
 				s.logger.Errorf("Failed to get target directory path for model %s: %v", modelInfo, err)
 				return err
 			}
-			err = utils.Retry(s.downloadRetry, 100*time.Millisecond, func() error {
-				downloadErr := s.downloadModel(ctx, osUri, destPath, task)
-				if downloadErr != nil {
-					// Check if context was cancelled
-					if ctx.Err() != nil {
-						s.logger.Infof("Download cancelled for model %s: %v", modelInfo, ctx.Err())
-						return ctx.Err()
+			if matchedKey, reused := s.findReadyObjectStorageModelWithSamePath(ctx, task, baseModelSpec, destPath); reused {
+				s.logger.Infof("Reusing Ready same-path model artifact for %s/%s from %s at %s", namespace, name, matchedKey, destPath)
+			} else {
+				err = utils.Retry(s.downloadRetry, 100*time.Millisecond, func() error {
+					downloadErr := s.downloadModel(ctx, osUri, destPath, task)
+					if downloadErr != nil {
+						// Check if context was cancelled
+						if ctx.Err() != nil {
+							s.logger.Infof("Download cancelled for model %s: %v", modelInfo, ctx.Err())
+							return ctx.Err()
+						}
+						s.logger.Errorf("Failed to download model %s (attempt %d/%d): %v",
+							modelInfo, s.downloadRetry, s.downloadRetry, downloadErr)
 					}
-					s.logger.Errorf("Failed to download model %s (attempt %d/%d): %v",
-						modelInfo, s.downloadRetry, s.downloadRetry, downloadErr)
-				}
-				return downloadErr
-			})
-			if err != nil {
-				s.logger.Errorf("All download attempts failed for model %s: %v", modelInfo, err)
+					return downloadErr
+				})
+				if err != nil {
+					s.logger.Errorf("All download attempts failed for model %s: %v", modelInfo, err)
 
-				// Record download failure in metrics
-				errorType := "download_error"
-				if strings.Contains(err.Error(), "MD5") {
-					errorType = "md5_verification_error"
-				}
-				s.metrics.RecordFailedDownload(modelType, namespace, name, errorType)
+					// Record download failure in metrics
+					errorType := "download_error"
+					if strings.Contains(err.Error(), "MD5") {
+						errorType = "md5_verification_error"
+					}
+					s.metrics.RecordFailedDownload(modelType, namespace, name, errorType)
 
-				s.markModelOnNodeFailed(task)
-				return err
+					s.markModelOnNodeFailed(task)
+					return err
+				}
 			}
 			// Parse model config and update ConfigMap
 			// We can pass either BaseModel or ClusterBaseModel based on the task's model type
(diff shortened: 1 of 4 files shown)


pkg/modelagent/gopher.go
```diff
@@ -736,6 +741,108 @@ func (s *Gopher) createOCIOSDataStore(baseModelSpec v1beta1.BaseModelSpec) (*oci
...
+func isReadyModelEntry(dataEntry string) bool {
+	var entry ModelEntry
+	if err := json.Unmarshal([]byte(dataEntry), &entry); err != nil {
+		return false
+	}
+	return entry.Status == ModelStatusReady
+}
+
+func sameModelStoragePath(currentStorage *v1beta1.StorageSpec, candidateStorage *v1beta1.StorageSpec, modelRootDir string, destPath string) bool {
+	if currentStorage == nil || candidateStorage == nil || currentStorage.Path == nil || candidateStorage.Path == nil {
```
truddy0, line 812: do we need to handle the case where the getDestPath is empty?

pkg/modelagent/gopher.go
```diff
@@ -736,6 +741,108 @@ func (s *Gopher) createOCIOSDataStore(baseModelSpec v1beta1.BaseModelSpec) (*oci
...
+		} else {
+			s.logger.Warnf("Cannot list ClusterBaseModels for same-path model reuse: %v", err)
+		}
+	}
+
+	if s.baseModelLister != nil {
+		baseModels, err := s.baseModelLister.List(labels.Everything())
+		if err == nil {
+			for _, model := range baseModels {
+				key := constants.GetModelConfigMapKey(model.Namespace, model.Name, false)
```
truddy0, line 790: nit: maybe we can extract reusable part between cluster basemodel and basemodel.

pkg/modelagent/gopher.go
```diff
@@ -736,6 +741,108 @@ func (s *Gopher) createOCIOSDataStore(baseModelSpec v1beta1.BaseModelSpec) (*oci
...
+	}
+	if !sameStringMapPtr(currentStorage.Parameters, candidateStorage.Parameters) {
+		return false
+	}
+
+	candidateSpec := v1beta1.BaseModelSpec{Storage: candidateStorage}
+	return getDestPath(&candidateSpec, modelRootDir) == destPath
+}
+
+func sameStringPtr(left *string, right *string) bool {
```
truddy0: these 2 funcs:sameStringPtr,  sameStringMapPtr should be moved to utils.go file as they are pure string validation methods

pkg/modelagent/gopher.go
```diff
@@ -736,6 +741,108 @@ func (s *Gopher) createOCIOSDataStore(baseModelSpec v1beta1.BaseModelSpec) (*oci
 	return ociOSDS, nil
 }
 
+func (s *Gopher) findReadyObjectStorageModelWithSamePath(ctx context.Context, task *GopherTask, baseModelSpec v1beta1.BaseModelSpec, destPath string) (string, bool) {
```
truddy0, line 749: let's add a comment for this func

pallasathena92:
we have this before https://github.com/ome-projects/ome/commit/f8d263f58a89955fed79694cb79ca43ad4e95155
which enable HuggingFace to leverage sha for artifact reuse.
we can use it in oci object storage as well

pkg/modelagent/gopher.go
op109lvb: No need to add.
`getDestPath` already falls back to `modelRootDir + "/" + storageUri` when `storage.path` is empty, and reuse still requires `os.Stat(destPath)` to pass.

pkg/modelagent/gopher.go
op109lvb: Thanks. I kept the BaseModel and ClusterBaseModel lookup loops separate because that matches the existing repo pattern: they are different CR types with different listers and namespace/key semantics. The shared comparison logic is already extracted into sameModelStoragePath, so I’d avoid a broader refactor in this focused PR.

op109lvb:
> we have this before [f8d263f](https://github.com/ome-projects/ome/commit/f8d263f58a89955fed79694cb79ca43ad4e95155) which enable HuggingFace to leverage sha for artifact reuse. we can use it in oci object storage as well

For the existing HF reuse path: It works because HF has a model-level commit SHA that model-agent can fetch before download and compare with the SHA recorded in the node ConfigMap. OCI Object Storage does not have the same model-level SHA in the current implementation. 

To prove an existing OCI local copy matches, the current path has to do per-object size/MD5 validation, which is the expensive work this PR is trying to skip for copied CRs. 
So I kept this PR scoped to OCI same-path Ready reuse: if the storage identity and resolved local destination are exactly the same and already Ready on this node, we skip downloadModel and avoid re-running per-file validation.

pallasathena92 (approved):

truddy0 (approved):

slin1237 (approved):

beiguo218:
My concern is that the current implementation also applies this reuse to `DownloadOverride`. That creates a redownload/refresh regression when two model CRs point to the same OCI storage identity and resolved local destination.

op109lvb:
> My concern is that the current implementation also applies this reuse to `DownloadOverride`. That creates a redownload/refresh regression when two model CRs point to the same OCI storage identity and resolved local destination.

Good point. The same-path reuse optimization is intended for the copied-CR latency path, not for refresh/update flows.

I checked the model-agent flow again: `DownloadOverride` can be generated for non-HF models when spec/label/annotation changes, so applying this reuse there could skip an intended refresh/validation for OCI models.

I’ll address this in a follow-up PR by scoping same-path OCI reuse to normal `Download` tasks only, and keeping `DownloadOverride` on the existing download/validation path.

PR #633, opened 2026-06-22
Add Kimi K2.5 generation vLLM runtime and K2.6/K2.7-Code models

Juno13340:
## What this PR does

Adds OME config for the Kimi K2.5-generation (`KimiK25ForConditionalGeneration`) models:

- A single shared vLLM `ClusterServingRuntime` `vllm-kimi-k25-single-node-8gpu` (single node, 8 GPU, TP=8) that both models auto-select.
- Two `ClusterBaseModel` entries: `kimi-k2-6` (`moonshotai/Kimi-K2.6`) and `kimi-k2-7-code` (`moonshotai/Kimi-K2.7-Code`).
- Registers all three in the respective `kustomization.yaml`.

## Why we need it

Enables serving the new Kimi K2.6 and K2.7-Code models on OME. Both share an identical vLLM serving config (same architecture, size range, args, images), so they are served by one generic runtime via `autoSelect` rather than duplicating per-model runtimes — consistent with how the Qwen3 guard/base runtimes were consolidated. The `k25` naming keeps it distinct from the existing original-K2 runtimes, which use `DeepseekV3ForCausalLM`.

Fixes #

## How to test

N/A — config-only change. Validated with `kubectl kustomize config/runtimes` and `kubectl kustomize config/models` (both build cleanly). After applying, a Kimi K2.6/K2.7-Code model auto-selects `vllm-kimi-k25-single-node-8gpu` without naming it in the InferenceService.

## Checklist

- [ ] Tests added/updated (if applicable) — N/A (config only)
- [ ] Docs updated (if applicable) — N/A
- [x] `kubectl kustomize` builds locally for both config/runtimes and config/models

### DIFF
--- a/config/models/moonshotai/Kimi-K2.6.yaml
+++ b/config/models/moonshotai/Kimi-K2.6.yaml
@@ -1,0 +1,14 @@
+apiVersion: ome.io/v1beta1
+kind: ClusterBaseModel
+metadata:
+  name: kimi-k2-6
+spec:
+  modelCapabilities:
+    - IMAGE_TEXT_TO_TEXT
+  vendor: moonshotai
+  displayName: moonshotai.kimi-k2-6
+  disabled: false
+  version: "1.0.0"
+  storage:
+    storageUri: hf://moonshotai/Kimi-K2.6
+    path: /raid/models/moonshotai/Kimi-K2.6
--- a/config/models/moonshotai/Kimi-K2.6.yaml
+++ b/config/models/moonshotai/Kimi-K2.6.yaml
@@ -4,6 +4,7 @@ metadata:
   name: kimi-k2-6
 spec:
   modelCapabilities:
+    - TEXT_TO_TEXT
     - IMAGE_TEXT_TO_TEXT
   vendor: moonshotai
   displayName: moonshotai.kimi-k2-6
--- a/config/runtimes/vllm/moonshotai/kimi-k25-single-node-8gpu-rt.yaml
+++ b/config/runtimes/vllm/moonshotai/kimi-k25-single-node-8gpu-rt.yaml
@@ -3,16 +3,12 @@ kind: ClusterServingRuntime
 metadata:
   name: vllm-kimi-k25-single-node-8gpu
   annotations:
-    ome.io/fit-quantization: int4
     ome.io/engine-ready-timeout-sec: "3600"
 spec:
   acceleratorRequirements:
     acceleratorClasses:
       - nvidia-b200-8
       - nvidia-h200-8
-      # - nvidia-b300-8
-      # - nvidia-h100-8
-      # - nvidia-a100-80gb-8
   disabled: false
   routerConfig:
     annotations:
@@ -28,6 +24,9 @@ spec:
         - containerPort: 8080
           name: http
       resources:
+        requests:
+          cpu: "1"
+          memory: 2Gi
         limits:
           cpu: "1"
           memory: 2Gi
@@ -75,7 +74,7 @@ spec:
         timeoutSeconds: 10
       startupProbe:
         httpGet:
-          path: /readiness
+          path: /health
           port: 8080
         failureThreshold: 177
         periodSeconds: 20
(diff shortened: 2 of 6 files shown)


config/models/moonshotai/Kimi-K2.6.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: ome.io/v1beta1
+kind: ClusterBaseModel
+metadata:
+  name: kimi-k2-6
+spec:
+  modelCapabilities:
+    - IMAGE_TEXT_TO_TEXT
```
YouNeedCryDear, line 8: Does it cover the default TEXT_TO_TEXT?

config/runtimes/vllm/moonshotai/kimi-k25-single-node-8gpu-rt.yaml
```diff
@@ -0,0 +1,188 @@
+apiVersion: ome.io/v1beta1
+kind: ClusterServingRuntime
+metadata:
+  name: vllm-kimi-k25-single-node-8gpu
```
YouNeedCryDear: maybe just simplify the name to tp8?

config/runtimes/vllm/moonshotai/kimi-k25-single-node-8gpu-rt.yaml
```diff
@@ -0,0 +1,188 @@
...
+  annotations:
+    ome.io/fit-quantization: int4
```
YouNeedCryDear: Do not add random annotations to the runtime

config/runtimes/vllm/moonshotai/kimi-k25-single-node-8gpu-rt.yaml
```diff
@@ -0,0 +1,188 @@
...
+    ome.io/engine-ready-timeout-sec: "3600"
+spec:
+  acceleratorRequirements:
+    acceleratorClasses:
+      - nvidia-b200-8
+      - nvidia-h200-8
+      # - nvidia-b300-8
```
YouNeedCryDear: remove comments if not needed

config/runtimes/vllm/moonshotai/kimi-k25-tp8-rt.yaml
```diff
@@ -0,0 +1,188 @@
...
+      prometheus.io/scrape: 'true'
+    labels:
+      logging-forward: enabled
+    runner:
+      name: router
+      image: docker.io/lightseekorg/smg:1.5.0
+      ports:
+        - containerPort: 8080
+          name: http
+      resources:
```
YouNeedCryDear, line 26: add request

config/runtimes/vllm/moonshotai/kimi-k25-single-node-8gpu-rt.yaml
```diff
@@ -0,0 +1,188 @@
...
+      livenessProbe:
+        httpGet:
+          path: /liveness
+          port: 8080
+        failureThreshold: 5
+        periodSeconds: 30
+        timeoutSeconds: 10
+      startupProbe:
+        httpGet:
+          path: /readiness
```
YouNeedCryDear: use /health

config/runtimes/vllm/moonshotai/kimi-k25-tp8-rt.yaml
```diff
@@ -0,0 +1,188 @@
...
+        #   tensorParallelismOverride:
+        #     tensorParallelSize: 8
+        # nvidia-h100-8:
+        #   tensorParallelismOverride:
+        #     tensorParallelSize: 8
+        # nvidia-a100-80gb-8:
+        #   tensorParallelismOverride:
+        #     tensorParallelSize: 8
+  modelSizeRange:
+    min: 150B
```
YouNeedCryDear, line 101: How is this size calculated? Isn't it a 1T parameter model?

YouNeedCryDear (changes requested):

config/models/moonshotai/Kimi-K2.6.yaml
Juno13340: Good catch, no, IMAGE_TEXT_TO_TEXT doesn't implicitly cover TEXT_TO_TEXT, they're independent enum values with no containment logic. Since both models are general purpose (K2.6) and coding (K2.7-Code) models where text is the primary use case, same as how gemma-3-* and Mistral-Small-3.1 declare both. I've added TEXT_TO_TEXT alongside IMAGE_TEXT_TO_TEXT in 421852c. Confirmed via config.json that both are genuinely multimodal (vision_config present, image + video), so IMAGE_TEXT_TO_TEXT stays too.

config/runtimes/vllm/moonshotai/kimi-k25-tp8-rt.yaml
Juno13340: 1T logically, yes, but OME matches on the safetensors element count, which ignores dtype. These are int4-packed, so it comes out ~150–300B, not 1T. Tried 900–1100B first and autoSelect failed for this exact reason.

PR #647, opened 2026-07-01
Add support to compare deploymentMode in runtime auto-selection logic

beiguo218:
## Context

Runtime auto-selection currently selects compatible runtimes based on model format, accelerator requirements, and related compatibility checks. For models that have separate single-node and multi-node runtimes, the selected runtime also needs to respect the desired component deployment mode.

This PR adds generic deployment mode matching to runtime auto-selection so an InferenceService can distinguish between runtimes intended for different deployment modes.

**Changes:**
- Add component-level `ome.io/deploymentMode` compatibility checks to runtime matching.
- Compare deployment mode annotations between:
  - ISVC `spec.engine.annotations` and runtime `engineConfig.annotations`
  - ISVC `spec.decoder.annotations` and runtime `decoderConfig.annotations`
- Ignore ISVC top-level metadata `ome.io/deploymentMode` during runtime selection.
- Preserve backward compatibility by only filtering out a runtime when:
  - the ISVC component (engine/decoder) has a valid deployment mode annotation,
  - the corresponding runtime component also has a valid deployment mode annotation,
  - and the two values differ.
- Treat missing, empty, or invalid deployment mode annotations as unset, so they do not add filtering constraints.
- Surface deployment mode mismatch reasons through compatibility details.
- Add tests for engine and decoder deployment mode matching, mismatch rejection, top-level annotation ignoring, invalid/empty annotation handling, and selector behavior.

## Compatibility

This keeps existing runtimes selectable when they do not declare component-level deployment mode annotations. Existing ISVCs without component-level deployment mode annotations continue to follow the previous runtime selection behavior.

## Test
Tested in cluster

## Checklist

### DIFF
--- a/pkg/runtimeselector/matcher.go
+++ b/pkg/runtimeselector/matcher.go
@@ -7,7 +7,6 @@ import (
 	"strconv"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/api/equality"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"sigs.k8s.io/ome/pkg/apis/ome/v1beta1"
@@ -32,14 +31,6 @@ func (m *DefaultRuntimeMatcher) IsCompatible(runtime *v1beta1.ServingRuntimeSpec
 	// Quick checks first
 	if runtime.IsDisabled() {
 		return false, &RuntimeDisabledError{RuntimeName: runtimeName}
-	}
-
-	// Check the runtime provides configs for all non-empty ISVC components.
-	if ok, reason := m.compareRequiredComponents(runtime, isvc); !ok {
-		return false, &RuntimeComponentMismatchError{
-			RuntimeName: runtimeName,
-			Reason:      reason,
-		}
 	}
 
 	// Check accelerator class compatibility
@@ -90,12 +81,6 @@ func (m *DefaultRuntimeMatcher) GetCompatibilityDetails(runtime *v1beta1.Serving
 	// Check if runtime is disabled
 	if runtime.IsDisabled() {
 		report.IncompatibilityReasons = append(report.IncompatibilityReasons, "runtime is disabled")
-		return report, nil
-	}
-
-	// Check that the runtime provides configs for all non-empty ISVC components.
-	if ok, reason := m.compareRequiredComponents(runtime, isvc); !ok {
-		report.IncompatibilityReasons = append(report.IncompatibilityReasons, reason)
 		return report, nil
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/runtimeselector/matcher.go
```diff
@@ -298,6 +314,96 @@ func (m *DefaultRuntimeMatcher) compareAcceleratorClass(runtime *v1beta1.Serving
...
+
+func inferenceServiceDecoderDeploymentMode(isvc *v1beta1.InferenceService) (constants.DeploymentModeType, bool) {
+	if isvc == nil || isvc.Spec.Decoder == nil {
+		return "", false
+	}
+
+	return deploymentModeFromAnnotations(isvc.Spec.Decoder.Annotations)
+}
+
+func getRuntimeEngineDeploymentMode(runtime *v1beta1.ServingRuntimeSpec) (constants.DeploymentModeType, bool) {
```
pallasathena92: would you combine these two getRuntimeEngineDeploymentMode and getRuntimeEngineDeploymentMode. The two function definiation and args are same

tobyxuge (approved):

slin1237 (changes requested):
this seems wrong
If isvc has engine and decode, and the defined runtime only has engine, webhook should fail 
if isvc has engine and decode, and the auto selected runtime doesn't have decode, fail this at webhook too

pkg/runtimeselector/matcher.go
beiguo218: Updated, thanks @pallasathena92

beiguo218:
> this seems wrong If isvc has engine and decode, and the defined runtime only has engine, webhook should fail if isvc has engine and decode, and the auto selected runtime doesn't have decode, fail this at webhook too

Do you mean when we either do ValidateRuntime or SelectRuntime, we should check if runtime can provide the config required by ISVC? We need to do such check in runtime selection/validation logic and then the webhook can catch it, right?

beiguo218:
> > this seems wrong If isvc has engine and decode, and the defined runtime only has engine, webhook should fail if isvc has engine and decode, and the auto selected runtime doesn't have decode, fail this at webhook too
> 
> Do you mean when we either do ValidateRuntime or SelectRuntime, we should check if runtime can provide the config required by ISVC? We need to do such check in runtime selection/validation logic and then the webhook can catch it, right?

I just made the updates based on the above understanding, please check when you get a chance, thanks.

pallasathena92:
> this seems wrong
> If isvc has engine and decode, and the defined runtime only has engine, webhook should fail 
> if isvc has engine and decode, and the auto selected runtime doesn't have decode, fail this at webhook too
> 

in current engine and decoder merge logic. If runtime doesn't have engine or decoder but the isvc has, the mergedengine or mergeddecoder will merged by isvc.

beiguo218:
> > this seems wrong
> > If isvc has engine and decode, and the defined runtime only has engine, webhook should fail
> > if isvc has engine and decode, and the auto selected runtime doesn't have decode, fail this at webhook too
> 
> in current engine and decoder merge logic. If runtime doesn't have engine or decoder but the isvc has, the mergedengine or mergeddecoder will merged by isvc.

Ah you are right, we do allow the case no config in runtime and everthing passed from ISVC, since we have the logic to merge ISVC components specs with runtime components specs.

beiguo218:
> > > this seems wrong If isvc has engine and decode, and the defined runtime only has engine, webhook should fail if isvc has engine and decode, and the auto selected runtime doesn't have decode, fail this at webhook too
> > 
> > 
> > Do you mean when we either do ValidateRuntime or SelectRuntime, we should check if runtime can provide the config required by ISVC? We need to do such check in runtime selection/validation logic and then the webhook can catch it, right?
> 
> I just made the updates based on the above understanding, please check when you get a chance, thanks.

Based on the comment from @pallasathena92 that we do have the case to allow decoder in isvc while it is not set in runtime due to whatever in isvc will get merged into the runtime, I reverted my later updates, please let me know if you do not think it makes sense.

pallasathena92 (approved):

slin1237 (approved):

PR #649, opened 2026-07-02
Handle long LWS-derived label values

beiguo218:
## Why we need it
Long inference service component names can produce generated resource names or pod labels that exceed Kubernetes label value length limits.

## Changes
- Truncate the LWS leader pod `app` label to 63 characters.
  This prevents long component names from being copied directly into the leader pod’s `app` label value and violating Kubernetes label constraints.

- Limit generated LWS name to 35 characters.
  This keeps derived names safely below label-value limits after downstream controllers add their own prefixes, indexes, and hash suffixes. It also adds coverage for short and long component names, including common derived label values.

## Tests
Tested

## Checklist

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -662,12 +662,14 @@ func ModelConfigName(isvcName string) string {
 	return fmt.Sprintf("modelconfig-%s", isvcName)
 }
 
-func LWSName(isvcName string) string {
-	var maxLen = 50
-	if len(isvcName) > maxLen {
-		isvcName = isvcName[len(isvcName)-maxLen:]
-	}
-	return fmt.Sprintf("lws-%s", isvcName)
+func LWSName(componentName string) string {
+	const (
+		lwsNamePrefix = "lws-"
+		// Leave room for downstream prefixes/suffixes when this name is embedded in Pod label values.
+		maxLWSNameLength = 35
+	)
+
+	return TruncateNameWithPrefix(componentName, lwsNamePrefix, maxLWSNameLength)
 }
 
 func InferenceServiceHostName(name string, namespace string, domain string) string {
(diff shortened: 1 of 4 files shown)


pkg/constants/constants.go
```diff
@@ -662,12 +662,14 @@ func ModelConfigName(isvcName string) string {
...
-	var maxLen = 50
-	if len(isvcName) > maxLen {
-		isvcName = isvcName[len(isvcName)-maxLen:]
-	}
-	return fmt.Sprintf("lws-%s", isvcName)
+func LWSName(componentName string) string {
+	const (
+		lwsNamePrefix = "lws-"
+		// Leave room for downstream prefixes/suffixes when this name is embedded in Pod label values.
+		maxLWSNameLength = 35
```
pallasathena92, line 669: why maxLWSNameLength is 35?

pkg/constants/constants.go
```diff
@@ -952,6 +954,19 @@ func TruncateNameWithMaxLength(name string, maxLength int) string {
 	return truncateWithHashTweaks(name, maxLength)
 }
 
+func TruncateNameWithPrefix(name, prefix string, maxLength int) string {
+	if maxLength <= 0 {
+		return ""
+	}
+
+	payloadMaxLength := maxLength - len(prefix)
```
pallasathena92, line 962: `lwsNamePrefix = "lws-"` define the prefix. is this possible payloadmaxlength <= 0?

pallasathena92: i see i see, we create this function for general purpose. even in this case, we define the prefix as "lws-", the next api call might have a longer prefix. it is better that we have this check here

pallasathena92 (approved):

slin1237 (approved):

pkg/constants/constants.go
beiguo218: Yes, when the passed prefix is super long which is longer than the maxLength, we need to do some special handling.

pkg/constants/constants.go
beiguo218: 35 is a conservative compromise: the LWS name may be embedded into downstream generated labels with auto-added prefixes, indexes, and hash suffixes, so we keep it well below the 63-character label value limit to leave enough buffer for those additions.

PR #660, opened 2026-07-10
Ignore parsed model metadata for DownloadOverride decisions

op109lvb:
## What this PR does

Updates model-agent Scout `DownloadOverride` detection to ignore parser-populated model metadata fields, while still triggering refresh for storage changes, label/annotation changes, and TensorRT-LLM download routing changes.

## Why we need it

Model config parsing can enrich `BaseModel` / `ClusterBaseModel` spec with metadata such as model type, framework, format, capabilities, and decoded configuration. Those metadata-only updates should not cause a `DownloadOverride`, because they do not change what artifact is downloaded.

Label/annotation updates are still preserved as a manual refresh path, so operators can force download validation and repair corrupted local files via the existing MD5/size check.

Fixes #

## How to test

- `go test ./pkg/modelagent -count=1`
- `go test ./pkg/modelagent -count=1 -run 'TestUpdateBaseModel_NoDuplicateTaskOnPolicyOnlyChange|TestUpdateClusterBaseModel_NoDuplicateTaskOnPolicyOnlyChange|TestGenerateDownloadOverride'`

## Checklist

### DIFF
--- a/pkg/modelagent/scout.go
+++ b/pkg/modelagent/scout.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"go.uber.org/zap"
 	v1 "k8s.io/api/core/v1"
@@ -45,6 +46,17 @@ type TensorRTLLMShapeFilter struct {
 	IsTensorrtLLMModel bool
 	ShapeAlias         string
 	ModelType          string
+}
+
+type downloadOverrideChangeCandidate struct {
+	name     string
+	old, new interface{}
+}
+
+type modelDownloadInputs struct {
+	Storage              *v1beta1.StorageSpec
+	IsTensorRTLLMModel   bool
+	TensorRTLLMModelType string
 }
 
 func NewScout(ctx context.Context, nodeName string,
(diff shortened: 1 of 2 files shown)


pkg/modelagent/scout.go
```diff
@@ -346,22 +358,15 @@ func (w *Scout) updateBaseModel(old, new interface{}) {
...
-		{"Spec", oldBaseModel.Spec, newBaseModel.Spec},
-	} {
-		result, err := kmp.SafeDiff(diff.old, diff.new, ignoreDownloadPolicy)
-		if err != nil {
-			w.logger.Errorf("Failed to diff %s for BaseModel: %s in namespace %s",
-				diff.name, newBaseModel.Name, newBaseModel.Namespace)
-			return
-		}
-		hasChanges = hasChanges || (result != "")
+		{"DownloadInputs", modelDownloadInputsFromSpec(oldBaseModel.Spec), modelDownloadInputsFromSpec(newBaseModel.Spec)},
```
pallasathena92: Can we use a clearer naming convention?

op109lvb: you mean `DownloadInputs`?

op109lvb: What about `downloadOverrideInputsFromSpec`

pallasathena92 (approved):

PR #662, opened 2026-07-11
[Core] Prioritize real downloads over validation replay

op109lvb:
## What this PR does

Splits model-agent normal work into separate download and validation-only FIFO lanes, then places already-Ready OCI models with existing local artifacts behind real download work.

It also snapshots models that were already Ready on this node before reconciliation marks restart replay tasks as Updating, and adds tests for queue ordering, delete superseding, startup classification, and demotion behavior.

## Why we need it

After model-agent restart, many already-downloaded models can be replayed as download tasks just to revalidate local artifacts. Those validation-heavy tasks can block real downloads for new models.

This change keeps real downloads ahead of validation replay without moving real download work onto the high-priority worker.

Fixes #

## How to test
- `go test -count=1 ./pkg/modelagent`
- E2E tests

## Checklist

### DIFF
--- a/pkg/modelagent/gopher.go
+++ b/pkg/modelagent/gopher.go
@@ -46,7 +46,7 @@ type GopherTask struct {
 	TensorRTLLMShapeFilter *TensorRTLLMShapeFilter
 	SamePathWaitStartedAt  time.Time
 	NormalPriorityOnly     bool
-	NormalValidationOnly   bool
+	RevalidationReplay     bool
 }
 
 type activeDownload struct {
@@ -85,8 +85,9 @@ type Gopher struct {
 const (
 	BigFileSizeInMB = 200
 
-	defaultSamePathWaitDelay   = 30 * time.Second
-	defaultSamePathWaitTimeout = 30 * time.Minute
+	defaultSamePathWaitDelay           = 30 * time.Second
+	defaultSamePathWaitTimeout         = 30 * time.Minute
+	defaultStartupReadySnapshotTimeout = 5 * time.Second
 )
 
 func NewGopher(
@@ -136,7 +137,9 @@ func NewGopher(
 }
 
 func (s *Gopher) Run(stopCh <-chan struct{}, numWorker int, numHighPriorityWorker int) {
-	s.captureStartupReadyModels(context.Background())
+	startupSnapshotCtx, cancelStartupSnapshot := context.WithTimeout(context.Background(), defaultStartupReadySnapshotTimeout)
+	defer cancelStartupSnapshot()
+	s.captureStartupReadyModels(startupSnapshotCtx)
 
 	// Start the ConfigMap reconciliation service
 	s.configMapReconciler.StartReconciliation()
@@ -198,7 +201,7 @@ func (s *Gopher) enqueueTask(task *GopherTask) {
 	if task.TaskType == Delete {
 		s.cancelActiveDownload(task)
 	} else {
-		s.markValidationOnlyIfStartupReady(task)
+		s.classifyStartupRevalidation(task)
 	}
 	s.taskQueue.enqueue(task)
 }
@@ -215,9 +218,6 @@ func (s *Gopher) runWorker() {
 		}
 		if task.TaskType == Delete {
 			s.cancelActiveDownload(task)
-		}
-		if s.deferStartupReadyValidationIfLocalPathExists(task) {
-			continue
 		}
 		err := s.processTask(task)
 		if err != nil {
@@ -689,29 +689,20 @@ func (s *Gopher) demoteToNormalPriority(task *GopherTask) {
 		return
 	}
 	task.NormalPriorityOnly = true
-	s.markValidationOnlyIfStartupReady(task)
+	s.classifyStartupRevalidation(task)
 	s.logger.Infof("Demoting %s to normal priority for fallback download/validation", getModelInfoForLogging(task))
 	s.enqueueTask(task)
 }
 
-func (s *Gopher) deferStartupReadyValidationIfLocalPathExists(task *GopherTask) bool {
-	if s.markValidationOnlyIfStartupReady(task) {
-		s.logger.Infof("Deferring MD5 validation for %s behind normal download work", getModelInfoForLogging(task))
-		s.enqueueTask(task)
-		return true
-	}
-	return false
-}
-
-func (s *Gopher) markValidationOnlyIfStartupReady(task *GopherTask) bool {
-	if task == nil || task.NormalValidationOnly || task.TaskType != Download {
+func (s *Gopher) classifyStartupRevalidation(task *GopherTask) bool {
+	if task == nil || task.RevalidationReplay || task.TaskType != Download {
 		return false
 	}
-	if !s.wasReadyAtStartupWithLocalPath(task) {
+	if !s.isStartupRevalidation(task) {
 		return false
 	}
 	task.NormalPriorityOnly = true
-	task.NormalValidationOnly = true
+	task.RevalidationReplay = true
 	return true
 }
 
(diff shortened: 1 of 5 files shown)


pkg/modelagent/gopher.go
```diff
@@ -679,10 +689,82 @@ func (s *Gopher) demoteToNormalPriority(task *GopherTask) {
...
+	for key, data := range configMap.Data {
+		if hasModelEntryStatus(data, ModelStatusReady) {
+			readyModelKeys[key] = struct{}{}
+		}
+	}
+	s.startupReadyModelKeys = readyModelKeys
+	s.logger.Infof("Captured %d Ready models from startup ConfigMap snapshot", len(readyModelKeys))
+}
+
+func (s *Gopher) wasReadyAtStartupWithLocalPath(task *GopherTask) bool {
```
EdHasNoLife: Can we add some more tests for this? The tests only ever seed startupReadyModelKeys manually with getModelID(...) — none of them actually populate it end-to-end via a real ConfigMap and then look it up via getModelID.

pkg/modelagent/gopher_test.go
```diff
@@ -1060,6 +1060,88 @@ func TestProcessTaskWithOptions_HighPriorityDemotesFallbackDownload(t *testing.T
 	assert.Equal(t, 0, g.taskQueue.len())
 }
 
+func TestDemoteToNormalPriorityClassifiesValidationBeforeEnqueue(t *testing.T) {
```
EdHasNoLife: Test setup bug in TestDemoteToNormalPriorityClassifiesValidationBeforeEnqueue. missingArtifact reuses the exact same storageURI and modelPath variables as alreadyDownloaded, so the directory for "missing-artifact" actually exists on disk too — the name is misleading, and the test only distinguishes the two cases via the startupReadyModelKeys membership, not via an actually-missing local path. It'd be clearer (and better coverage) to give missingArtifact its own non-existent path.

pkg/modelagent/gopher.go
op109lvb: Added an end-to-end test that seeds a real ConfigMap, captures startup Ready entries, and verifies the captured key is used through getModelID to classify the download as validation-only.

pkg/modelagent/gopher_test.go
op109lvb: Good catch! Fixed the fixture so missingArtifact uses its own nonexistent local path, with an assertion that the path is absent before enqueueing.

pkg/modelagent/gopher.go
```diff
@@ -133,6 +136,8 @@ func NewGopher(
 }
 
 func (s *Gopher) Run(stopCh <-chan struct{}, numWorker int, numHighPriorityWorker int) {
+	s.captureStartupReadyModels(context.Background())
```
pallasathena92: captureStartupReadyModels(context.Background()) does a live ConfigMaps().Get() as the first statement of Run() — the reconciler, dispatcher, and every worker wait on it. The failure path already degrades gracefully to an empty snapshot (today's behavior), so a context.WithTimeout is free insurance against a throttled API server stalling startup. Relatedly, on a cold node the ConfigMap doesn't exist yet, so getConfigMap logs at Errorf and this logs Warnf — an alarming pair for a normal first boot. Consider special-casing apierrors.IsNotFound.

pkg/modelagent/gopher.go
```diff
@@ -209,6 +216,9 @@ func (s *Gopher) runWorker() {
 		if task.TaskType == Delete {
 			s.cancelActiveDownload(task)
 		}
+		if s.deferStartupReadyValidationIfLocalPathExists(task) {
```
pallasathena92: This classifier is redundant.
For task sits in a normal lane. the funtion markValidationOnlyIfStartupReady() in #201 already classified it on that enqueue. The deferStartupReadyValidationIfLocalPathExists call the same function  markValidationOnlyIfStartupReady() and it will return false because task.NormalValidationOnly is true.
For task demote from high lane. markValidationOnlyIfStartupReady in demoteToNormalPriority and markValidationOnlyIfStartupReady() in #201 already classified it during demote's re-enqueue.

pkg/modelagent/gopher.go
```diff
@@ -679,10 +689,82 @@ func (s *Gopher) demoteToNormalPriority(task *GopherTask) {
...
+
+	var baseModelSpec v1beta1.BaseModelSpec
+	if task.BaseModel != nil {
+		baseModelSpec = task.BaseModel.Spec
+	} else if task.ClusterBaseModel != nil {
+		baseModelSpec = task.ClusterBaseModel.Spec
+	} else {
+		return false
+	}
+	if baseModelSpec.Storage == nil || baseModelSpec.Storage.StorageUri == nil || baseModelSpec.Storage.Path == nil {
```
pallasathena92: Nit. wasReadyAtStartupWithLocalPath requires Storage.Path != nil but not non-empty. With Path = ptr(""), getDestPath falls back to modelRootDir + "/" + storageUri, which just fails os.Stat — safe, but the nil-check reads as though it's guarding the dest-path computation when it isn't. we can gate it here as well

pkg/modelagent/gopher.go
pallasathena92: isStartupRevalidation. how about this name?

pkg/modelagent/gopher.go
```diff
@@ -679,10 +689,82 @@ func (s *Gopher) demoteToNormalPriority(task *GopherTask) {
...
+func (s *Gopher) deferStartupReadyValidationIfLocalPathExists(task *GopherTask) bool {
+	if s.markValidationOnlyIfStartupReady(task) {
+		s.logger.Infof("Deferring MD5 validation for %s behind normal download work", getModelInfoForLogging(task))
+		s.enqueueTask(task)
+		return true
+	}
+	return false
+}
+
+func (s *Gopher) markValidationOnlyIfStartupReady(task *GopherTask) bool {
```
pallasathena92: classifyStartupRevalidation

pkg/modelagent/gopher.go
```diff
@@ -46,6 +46,7 @@ type GopherTask struct {
 	TensorRTLLMShapeFilter *TensorRTLLMShapeFilter
 	SamePathWaitStartedAt  time.Time
 	NormalPriorityOnly     bool
+	NormalValidationOnly   bool
```
pallasathena92: RevalidationReplay

pkg/modelagent/gopher.go
op109lvb: Addressed. Startup snapshot now uses a fixed 5s startup-only timeout, and NotFound is treated as a cold start with info level logging instead of warning/error noise.

pkg/modelagent/gopher.go
op109lvb: Addressed. Empty Storage.Path no longer qualifies for startup revalidation, and added a test to cover.

pkg/modelagent/gopher.go
op109lvb: Sounds good. Addressed.

pkg/modelagent/gopher.go
op109lvb: Addressed

pkg/modelagent/gopher.go
op109lvb: Addressed

pkg/modelagent/gopher.go
op109lvb: Good catch! Startup revalidation is now classified before enqueue, and the queue itself defers RevalidationReplay behind normal downloads, so the worker-side requeue wrapper is no longer needed.
Addressed

EdHasNoLife (approved):

frankzhouhr (approved):

pallasathena92 (approved):

beiguo218 (approved):

PR #664, opened 2026-07-13
[Core] Reuse OCI model artifacts by Hugging Face origin

op109lvb:
## What this PR does
This PR adds shared node-local artifact reuse for HF-origin OCI Object Storage models by using HF model ID + commit SHA as the artifact identity. It downloads once to a canonical parent, symlinks model-local paths to that parent, tracks parent/child relationships, and adds restart/retry handling so shared parents are recovered or validated safely after model-agent restarts.

### More details

Adds node-local reuse for HF-origin OCI Object Storage artifacts when a model CR provides HF model ID, HF commit SHA, `ReuseIfExists`, and a SHA-scoped OCI Object Storage URI.

Model-agent now treats `<hf-model-id>/<sha>` as the shared artifact identity:
- Downloads the artifact once into `_artifacts/<hf-model-id>/<sha>`.
- Creates each model-local path as a symlink to that shared parent.
- Records parent/child relationships in the node ConfigMap.
- Avoids reusing legacy/self-parent entries that do not have shared parent metadata.

It also adds lifecycle handling for the shared parent:
- Deleting one child only removes that child symlink.
- Deleting the last child removes the shared parent when no references remain.
- `DownloadOverride` repairs/redownloads the shared parent, since all children point to the same files.

Restart and concurrency safety is included:
- Uses the synthetic HF parent ConfigMap `Updating` state to ensure only one worker validates or rebuilds a shared parent at a time.
- On model-agent startup, recovers synthetic HF parent entries left in `Updating`: if the ready marker exists, marks the parent `Ready`; if missing, marks it `Failed` so later child tasks can rebuild.
- Validates each Ready HF parent once after model-agent restart.
- Retries recovery/validation on transient ConfigMap read/update failures.

Legacy behavior is preserved for model CRs that do not provide the HF reuse metadata. Backfill is out of scope.

## Why we need it

Multiple model CRs can point to the same HF model revision after the artifact has been copied into OCI Object Storage. Reusing the node-local artifact by HF model ID + SHA avoids duplicate downloads and saves node disk space, while still preserving repair behavior for corrupted files.

Fixes #

## How to test
```
go test ./pkg/modelagent ./pkg/modelparser
```

Run model-agent unit tests covering HF parent path selection, symlink reuse, child tracking, delete cleanup, startup recovery, startup validation, and DownloadOverride repair behavior.

E2E: create two model CRs with the same HF model ID + SHA, `ReuseIfExists`, and the same SHA-scoped OCI Object Storage URI. Verify the first model downloads to `_artifacts/<hf-model-id>/<sha>` and the second model-local path is a symlink to the same parent.

E2E coverage:
- First CR with HF model ID + SHA downloads OCI Object Storage artifact to `_artifacts/<hf-model-id>/<sha>`.
- Second CR with same HF model ID + SHA creates a symlink to the same shared parent.
- Delete one child CR: shared parent remains.
- Delete last child CR: shared parent is cleaned up.
- Existing legacy CRs without shared HF metadata keep old behavior.
- Restart model-agent: stale `Updating` parent entries recover or fail safely.
- Restart model-agent: Ready HF parent is validated once before reuse.
- Corrupted shared parent: `DownloadOverride` redownloads the parent artifact.
- ClusterBaseModel and BaseModel both preserve expected DownloadOverride behavior.

## Checklist

### DIFF
--- a/pkg/modelagent/configmap_reconciler.go
+++ b/pkg/modelagent/configmap_reconciler.go
@@ -26,10 +26,6 @@ const (
 	ConfigAttr        = "config"
 	ArtifactAttr      = "artifact"
 	ShaAttr           = "sha"
-	OriginAttr        = "origin"
-	OriginTypeAttr    = "type"
-	HFModelIDAttr     = "hfModelId"
-	HFCommitSHAAttr   = "hfCommitSha"
 	ParentPathAttr    = "parentPath"
 	ChildrenPathsAttr = "childrenPaths"
 )
@@ -1047,100 +1043,6 @@ func (c *ConfigMapReconciler) FindMatchedModelFromConfigMap(configMap *corev1.Co
 	return matchedParentName, matchedParentPath, searchingError
 }
 
-/*
-FindMatchedModelFromConfigMapByArtifactIdentity scans Ready entries in the node
-ConfigMap for an artifact that has the same provenance identity. This is used
-for OCI artifacts that were originally imported from Hugging Face, where
-config.artifact.origin carries the HF model ID and resolved commit SHA.
-
-Returns the parent model key and parent artifact path for the best match,
-preferring the parent that already has the most children so new children keep
-pointing at the established parent artifact.
-*/
-func (c *ConfigMapReconciler) FindMatchedModelFromConfigMapByArtifactIdentity(configMap *corev1.ConfigMap, identity ArtifactIdentity, modelType string, currentModelTypeAndNodeName string) (string, string, error) {
-	var searchingError error
-
-	var matchedParentName, matchedParentPath string
-	var matchedParentChildrenNum = -1
-	for modelKey, jsonStr := range configMap.Data {
-		if !strings.HasPrefix(strings.ToLower(modelKey), strings.ToLower(modelType)) {
-			continue
-		}
-		if strings.EqualFold(modelKey, currentModelTypeAndNodeName) {
-			continue
-		}
-
-		var obj map[string]interface{}
-		if err := json.Unmarshal([]byte(jsonStr), &obj); err != nil {
-			searchingError = fmt.Errorf("fail to Unmarshal %s during FindMatchedModelFromConfigMapByArtifactIdentity: %s", jsonStr, err)
-			c.logger.Errorf(searchingError.Error())
-			continue
-		}
-		status, ok := obj["status"].(string)
-		if !ok || !strings.EqualFold(status, string(ModelStatusReady)) {
-			continue
-		}
-		config, ok := obj[ConfigAttr].(map[string]interface{})
-		if !ok {
-			continue
-		}
-		artifact, ok := config[ArtifactAttr].(map[string]interface{})
-		if !ok || !artifactMatchesIdentity(artifact, identity) {
-			continue
-		}
-
-		rawParent, ok := artifact[ParentPathAttr].(map[string]interface{})
-		if !ok {
-			searchingError = fmt.Errorf("parentPath is not an object")
-			continue
-		}
-		if len(rawParent) != 1 {
-			searchingError = fmt.Errorf("expected exactly one parentPath entry, got %d", len(rawParent))
-			continue
-		}
-
-		for parentName, value := range rawParent {
-			parentPath, ok := value.(string)
-			if !ok {
-				searchingError = fmt.Errorf("parentPath value for %q is not a string", parentName)
-				continue
-			}
-			if strings.EqualFold(parentName, currentModelTypeAndNodeName) {
-				continue
-			}
-
-			childrenNum := len(extractChildrenPaths(artifact))
-			if childrenNum > matchedParentChildrenNum {
-				matchedParentChildrenNum = childrenNum
-				matchedParentName = parentName
-				matchedParentPath = parentPath
-			}
-		}
-	}
-	c.logger.Infof("matched artifact identity parentName: %s, matchedParentPath: %s", matchedParentName, matchedParentPath)
-	return matchedParentName, matchedParentPath, searchingError
-}
-
-func artifactMatchesIdentity(artifact map[string]interface{}, identity ArtifactIdentity) bool {
-	origin, ok := artifact[OriginAttr].(map[string]interface{})
-	if !ok {
-		return false
-	}
-	originType, ok := origin[OriginTypeAttr].(string)
-	if !ok || !strings.EqualFold(originType, identity.OriginType) {
-		return false
-	}
-	hfModelID, ok := origin[HFModelIDAttr].(string)
-	if !ok || hfModelID != identity.HFModelID {
-		return false
-	}
-	hfCommitSHA, ok := origin[HFCommitSHAAttr].(string)
-	if !ok {
-		return false
-	}
-	return strings.EqualFold(hfCommitSHA, identity.HFCommitSHA)
-}
-
 func extractChildrenPaths(artifact map[string]interface{}) []string {
 	// Read childrenPaths leniently and convert to []string
 	children := make([]string, 0)
@@ -1177,21 +1079,6 @@ func (c *ConfigMapReconciler) getModelDataByArtifactSha(ctx context.Context, tar
 		return "", "", fmt.Errorf("failed to get ConfigMap %s: %v", c.nodeName, err)
 	}
 	return c.FindMatchedModelFromConfigMap(cm, targetSha, modelType, currentModelTypeAndNodeName)
-}
-
-// getModelDataByArtifactIdentity fetches the node ConfigMap and searches for a
-// Ready entry with the same normalized artifact identity.
-func (c *ConfigMapReconciler) getModelDataByArtifactIdentity(ctx context.Context, identity ArtifactIdentity, modelType string, currentModelTypeAndNodeName string) (string, string, error) {
-	cm, err := c.kubeClient.CoreV1().ConfigMaps("ome").Get(ctx, c.nodeName, metav1.GetOptions{})
-	if err != nil {
-		if errors.IsNotFound(err) {
-			c.logger.Warn("cannot find configmap %s", c.nodeName)
-			return "", "", fmt.Errorf("cannot find configmap %s", c.nodeName)
-		}
-		c.logger.Errorf("Failed to get ConfigMap %s: %v", c.nodeName, err)
-		return "", "", fmt.Errorf("failed to get ConfigMap %s: %v", c.nodeName, err)
-	}
-	return c.FindMatchedModelFromConfigMapByArtifactIdentity(cm, identity, modelType, currentModelTypeAndNodeName)
 }
 
 // addPathToChildrenPaths appends newPath to the config.artifact.childrenPaths array if the newPath is not contained in the children paths
--- a/pkg/modelagent/configmap_reconciler.go
+++ b/pkg/modelagent/configmap_reconciler.go
@@ -26,6 +26,10 @@ const (
 	ConfigAttr        = "config"
 	ArtifactAttr      = "artifact"
 	ShaAttr           = "sha"
+	OriginAttr        = "origin"
+	OriginTypeAttr    = "type"
+	HFModelIDAttr     = "hfModelId"
+	HFCommitSHAAttr   = "hfCommitSha"
 	ParentPathAttr    = "parentPath"
 	ChildrenPathsAttr = "childrenPaths"
 )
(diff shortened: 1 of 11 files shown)


pkg/modelagent/hf_artifact_identity.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	sanitized := strings.Trim(b.String(), ".")
+	if sanitized == "" {
+		return "unknown"
+	}
+	return sanitized
+}
+
+func canonicalHuggingFaceArtifactPath(destPath string, identity ArtifactIdentity) string {
+	return filepath.Join(filepath.Dir(destPath), constants.ModelArtifactsDirectory, filepath.FromSlash(strings.Trim(strings.TrimSpace(identity.HFModelID), "/")), strings.ToLower(identity.HFCommitSHA))
```
EdHasNoLife, line 184: shall we add sanitization against `../` sequences

pkg/modelagent/hf_artifact_identity.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// BaseModel downloads can have model-local child paths under a local store, so
+// place shared HF parents in that same store. ClusterBaseModel keeps the
+// existing model-root layout for backward compatibility.
+func canonicalHuggingFaceArtifactPathForTask(task *GopherTask, modelRootDir string, destPath string, identity ArtifactIdentity) string {
+	if task != nil && task.BaseModel != nil {
+		return canonicalHuggingFaceArtifactPath(destPath, identity)
+	}
+	return filepath.Join(modelRootDir, filepath.FromSlash(strings.Trim(strings.TrimSpace(identity.HFModelID), "/")), strings.ToLower(identity.HFCommitSHA))
```
EdHasNoLife, line 194: Ditto

pkg/modelagent/hf_artifact_identity.go
op109lvb: HF model IDs should not contain those sequences per Hugging Face repo ID rules, and upstream should already validate that. I still addressed this by adding model-agent-side validation as defense-in-depth before using the annotation to build local artifact paths.

pkg/modelagent/hf_artifact_identity.go
op109lvb: Addressed

EdHasNoLife (approved):

frankzhouhr (approved):

pkg/modelagent/gopher.go
```diff
@@ -591,7 +727,7 @@ func (s *Gopher) processTaskWithOptions(task *GopherTask, allowFallbackDownload
 			s.logger.Infof("Starting deletion for model %s", modelInfo)
 			destPath := getDestPath(&baseModelSpec, s.modelRootDir)
 			// check if it needs to skip artifact deletion
-			isSkippingDeletion, _, _, _ := s.isSkippingArtifactDeletion(ctx, task, destPath, false)
+			isSkippingDeletion, isRemoveParent, parentName, parentDir := s.isSkippingArtifactDeletion(ctx, task, destPath, s.hasSharedArtifactMetadata(ctx, task))
```
pallasathena92: in function hasSharedArtifactMetadata().
```
exists, dataEntry, err := s.configMapReconciler.getDataEntryBasedOnModelKey(ctx, modelTypeAndModelName)
	if err != nil {
		s.logger.Warnf("cannot inspect artifact metadata for %s during deletion: %v", modelTypeAndModelName, err)
		return false
	}
```
The transient read failure will return false. 
in isSkippingArtifactDeletion(). The false value return from hasSharedArtifactMetadata() will make the isRemoveParent as false.
```
 if needsConsiderChildrenPath {                                                     
          childrenPaths, parentName, parentDir, parseErr := s.parseModelConfigDataEntry(...)
          hasChildren := hasChildrenPaths(childrenPaths, parseErr)
          s.removeChildPathFromParentConfigMapIfNecessary(...)   
          isRemoveParent := s.isRemoveParentArtifactDirectory(...)
          return hasChildren, isRemoveParent, parentName, parentDir
      } else {
          return hasReserveLabel, false, "", ""                 // 2020  ← gate==false: no detach, isRemoveParent=false
      }
```
However, the child model symlink is still deleted back at line 732.

it is a rare case, failing to true just makes isSkippingArtifactDeletion do one extra parse that finds no parent/children and no-ops

op109lvb: Good point, thank you. Parent deletion should not happen in this failure path because `hasSharedArtifactMetadata=false` makes `isRemoveParent=false`, so `deleteParentArtifactIfUnreferenced` becomes a no-op.

The real risk is stale parent metadata: the child symlink/local path could be deleted, then the child ConfigMap entry could be removed when the model is marked `Deleted`, but the synthetic parent still keeps that child path in `childrenPaths`. Since the delete task completes successfully, there is no guaranteed retry to clean that up later.

I agree we should fail conservatively here: if we cannot inspect shared artifact metadata due to a ConfigMap read error, we should not complete child deletion. We should requeue/return an error before deleting the local path or removing the child entry, then retry when the ConfigMap is readable.

pkg/modelagent/configmap_reconciler.go
```diff
@@ -1043,6 +1047,100 @@ func (c *ConfigMapReconciler) FindMatchedModelFromConfigMap(configMap *corev1.Co
...
+FindMatchedModelFromConfigMapByArtifactIdentity scans Ready entries in the node
+ConfigMap for an artifact that has the same provenance identity. This is used
+for OCI artifacts that were originally imported from Hugging Face, where
+config.artifact.origin carries the HF model ID and resolved commit SHA.
+
+Returns the parent model key and parent artifact path for the best match,
+preferring the parent that already has the most children so new children keep
+pointing at the established parent artifact.
+*/
+func (c *ConfigMapReconciler) FindMatchedModelFromConfigMapByArtifactIdentity(configMap *corev1.ConfigMap, identity ArtifactIdentity, modelType string, currentModelTypeAndNodeName string) (string, string, error) {
```
pallasathena92: do we really call this function? 
The matcher's only non-test caller is getModelDataByArtifactIdentity.
And getModelDataByArtifactIdentity has zero caller.

op109lvb: good catch, cleaned up, thanks

pkg/modelagent/gopher_hf_artifact_startup.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (s *Gopher) validateStartupHuggingFaceArtifactParentIfNeeded(ctx context.Context, task *GopherTask, parentKey string, identity ArtifactIdentity, downloadObjectStorageModel func(string) error) (bool, error) {
+	if strings.TrimSpace(parentKey) == "" {
+		return false, nil
+	}
+	var validationPath string
+	var acquired bool
+	s.startupHuggingFaceParentValidationMutex.Lock()
+	if s.claimStartupHuggingFaceArtifactParentValidationLocked(parentKey) {
```
pallasathena92, line 189: 1. Worker A runs validateStartupHuggingFaceArtifactParentIfNeeded:182 — claims the validation key, sees parent Ready, calls the function acquireHuggingFaceArtifactParentRebuild:191
2. Meanwhile worker B (repairgopher.go:588]) flips the parent to Updating first.
3. A's CAS fails: gopher_hf_artifact.go:123 → requeueSamePathInFlightReuseWait schedules task onto gopherChan via time.AfterFunc (enqueue # 1) → returns ("", false, nil).
4. Back in validate…: validationPath == "" → returns (stop=false, nil) at [line 201]. The if !acquired { return true, nil } at line 203 is unreachable — !acquired always implies validationPath == "". Dead code hiding the bug.
5. Caller continues: reuse fails (parent is Updating) → requeueIfHuggingFaceArtifactParentUpdating → requeueSamePathInFlightReuseWait again (enqueue # 2).
5. Both AfterFuncs fire ~30s later. [enqueue] has no dedup for Download tasks (only Delete runs removeSupersededTasks), so the queue holds the same pointer twice; two workers pop it concurrently.

Fix. Two-part: (a) remove the requeue side-effect from acquireHuggingFaceArtifactParentRebuild — return a tri-state (acquired / busy / timedOut) and let callers requeue; (b) in validate…, treat "CAS attempted and lost" as stop=true. Minimal patch: track casAttempted, and check if casAttempted && !acquired { return true, nil } before the validationPath == "" early-return.

pkg/modelagent/gopher.go
```diff
@@ -714,21 +843,34 @@ func (s *Gopher) captureStartupReadyModels(ctx context.Context) {
...
-			return
+		} else {
+			s.logger.Warnf("Cannot capture startup Ready model snapshot: %v", err)
 		}
-		s.logger.Warnf("Cannot capture startup Ready model snapshot: %v", err)
 		s.startupReadyModelKeys = map[string]struct{}{}
+		s.startupHuggingFaceParentValidationMutex.Lock()
+		s.startupHuggingFaceParentValidationKeys = map[string]struct{}{}
+		s.startupHuggingFaceParentValidationTried = map[string]struct{}{}
+		s.startupHuggingFaceParentSnapshotMissing = true
```
pallasathena92: 1. captureStartupReadyMode ls: on IsNotFound and on real errors both to → snapshotMissing = true.
2. Task builds a brand-new parent via reserve → download → Ready.
3. Next task with the same identity enters validate… → claimStartupHuggingFaceArtifactParentValidationLocked (gopher_hf_artifact_startup.go#L152): key not in (empty) snapshot set, but snapshotMissing is true and key not yet in Tried.
4. Parent is Ready → Updating → full validation replay (list bucket, MD5 every file) → mark Ready again.

So for new node, NotFound should mean "no parents existed, nothing to validate": set snapshotMissing = false on the IsNotFound branch, keep true only for genuine read failures (where Ready parents might exist unseen). Safe because any parent created after startup was just downloaded+checksummed — validation would be a no-op by construction.

pkg/modelagent/gopher_hf_artifact.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	artifact, err := s.linkHuggingFaceOriginArtifact(ctx, task, name, destPath, matchedModelKey, parentPath, identity)
+	if err != nil {
+		return nil, false, err
+	}
+	return artifact, true, nil
+}
+
+func (s *Gopher) repairHuggingFaceOriginArtifactParent(ctx context.Context, task *GopherTask, name string, destPath string,
```
pallasathena92, line 62: Repair (DownloadOverride, [repairHuggingFaceOriginArtifactParent:88]), rebuild, and startup validation all re-download into the live parent directory while children symlinks keep serving from it.
children keep their Ready status the entire time the parent is being repaired. There is a edge case, some child model serving update or restart, it will serve with a downloading model artifact.

pkg/modelagent/gopher_hf_artifact.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		s.logger.Warnf("cannot inspect Hugging Face artifact parent %s: %v", parentKey, err)
+		return "", "", "", false, err
+	}
+	if !exists {
+		return "", "", "", false, nil
+	}
+
+	var entry ModelEntry
+	if err := json.Unmarshal([]byte(dataEntry), &entry); err != nil {
```
pallasathena92, line 196: Corrupt parent entry = permanent failure loop — getHuggingFaceArtifactParentWithError treats unparseable entries as "no parent" (err=nil), sending the task to reserveHuggingFaceArtifactParentEntry, which errors on the same corrupt entry — retry forever until someone hand-edits the ConfigMap. Consider auto-quarantine/delete of corrupt synthetic entries.

in validation side, the code sawllow the error. and then reserveHuggingFaceArtifactParentEntry/
markUpdating / markReady / markFailed (writes) side trates it as hard error.
```
if existingDataEntry, exists := currentConfigMap.Data[parentName]; exists {
    if err := json.Unmarshal(...); err != nil {
        return false, currentConfigMap, fmt.Errorf("failed to parse existing ... %s: %w", parentName, err)
    }
    if existing.Config == nil || existing.Config.Artifact.Origin == nil {
        return ..., fmt.Errorf("... missing origin metadata")
    }
    if /* origin mismatch */ {
        return ..., fmt.Errorf("... mismatched origin metadata")
    }
```
then the task die silently. when the next time scout resync / CR touch / agent restart re-enqueues the task, the same loop will happen again and again.

In the reserve/setStatus closures, on unmarshal error || Config==nil || Origin==nil:

```
s.logger.Errorf("quarantining corrupt HF artifact parent entry %s: %v", parentName, err)
delete(currentConfigMap.Data, parentName)   // then fall through to the create-fresh branch
```
— i.e., treat unreadable synthetic entries as deletable state, since they're pure cache (the source of truth is the bucket + on-disk artifact, and the ready-marker/rebuild machinery already handles a vanished entry). Keep the origin-mismatch case a hard error (it's a real identity conflict, not corruption) but consider falling back to plain per-model download instead of failing the task.

pkg/modelagent/gopher_hf_artifact_startup.go
op109lvb: Good catch. Addressed. 
Made the parent acquire helper side-effect free and moved the wait/requeue into explicit caller handling. Startup validation now stops processing and requeues once when another worker already owns the parent rebuild. Added coverage for the acquire-busy case.

pkg/modelagent/gopher.go
op109lvb: Good catch. Addressed. 
`ConfigMap NotFound` is now treated as a known empty startup snapshot, not as a missing/unavailable snapshot. Only real ConfigMap read failures trigger best-effort startup parent validation.

pkg/modelagent/gopher_hf_artifact.go
op109lvb: I agree with the concern, will handle this in a follow-up PR because it changes the repair semantics rather than the reuse ownership/corrupt-metadata fixes in this PR. I’ll open a separate issue to harden shared parent repair so DownloadOverride does not expose partial rewrites to sibling models or active serving pods.

pkg/modelagent/gopher_hf_artifact.go
op109lvb: Thanks for catching this. Addressed. 
The synthetic HF parent reservation and DownloadOverride repair paths now self-heal corrupt or structurally invalid parent entries. Parseable invalid entries preserve existing parent path and children; valid mismatched origin metadata still fails. Added tests for corrupt JSON, missing metadata, DownloadOverride repair, and Updating ownership preservation.

pkg/modelagent/gopher_hf_artifact.go
op109lvb: https://github.com/ome-projects/ome/issues/685

PR #666, opened 2026-07-13
feat: add Nemotron video-to-text capability

caihui-lol:
## What this PR does

Adds the VIDEO_TEXT_TO_TEXT model capability.
Includes it in the Kubernetes API validation enum.
Maps it between model configuration and OME API capabilities.
Adds it to the NemotronH Nano capability classification rule.
Updates classification and capability-mapping tests.

## Why we need it

The imported model nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16 supports video-and-text input with text output, but OME currently omits VIDEO_TEXT_TO_TEXT from its detected capabilities.
This causes the imported model metadata to represent only its image, text, and audio capabilities.

## How to test

- go test
- E2E test by importing the omni model

## Checklist


slin1237:
Did u run make generate and make manifest?
I think it will update them too?

caihui-lol:
> Did u run make generate and make manifest? I think it will update them too?

thanks @slin1237 for reviewing. I ran both `make generate` and `make manifests`. `make manifests` produced no tracked changes. `make generate` produced no capability-related output - only unrelated checksum and whitespace-only generator noise, so I excluded them from this PR for review. The current generated schema represents modelCapabilities.items as an unconstrained string, so this enum addition does not change the CRD or Helm manifests.

pallasathena92 (approved):

pallasathena92 (changes requested):
thanks for contributing, please resolve the conflict and request review again.

caihui-lol:
thanks @pallasathena92 ! rebase with [PR#672](https://github.com/ome-projects/ome/pull/672) to resolve the conflicts. as PR#672 added a more granular capability AUDIO_TEXT_TO_TEXT to distinguish with AUDIO_TO_TEXT, also update the nemotron rule to use AUDIO_TEXT_TO_TEXT instead. requesting review again.

pkg/modelconfig/capability.go
beiguo218, line 169: I am thinking a model type containing "nemotronh_nano" does not necessarily support
audio or video input modalities.

like nvidia/NVIDIA-Nemotron-Nano-12B-v2-VL-BF16, it does not support audio input

beiguo218, line 187: For nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-BF16 model, it seems it should go through this omniRule, while you may need to update it to add audio output modality conditionally.

beiguo218 (changes requested):

pkg/modelconfig/capability.go
beiguo218: I feel that, if we can make `omniRule` more precise, it should be placed before `nemotronHNanoRule` since capability rules use first-match-wins semantics. This would allow Omni models to receive their complete, model-specific capability set, while `nemotronHNanoRule` remains a fallback for other NemotronH Nano variants.

caihui-lol: yeah this is a legit concern. i just audited the nvidia collections, the [VL BF16](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-12B-v2-VL-BF16), [VL FP8](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-12B-v2-VL-FP8), and [VL NVFP4](https://huggingface.co/nvidia/NVIDIA-Nemotron-Nano-12B-v2-VL-NVFP4-QAD) checkpoints use `NemotronH_Nano_VL_V2` type and do not support audio. 

One quick fix i think we may need a stricter rule the nemotron nano omni model like a full match `NemotronH_Nano_Omni_Reasoning_V3`. the omniRule below is not legit for this model either as its capability is multimodal input to text. wdyt?

beiguo218:
We decided to make the current changes as simple as possible:
1. Just in nemotronHNanoRule, instead of directly adding CapabilityAudioTextToText and CapabilityVideoTextToText, we do a further check for the model type - only add those 2 capabilities when it comes to nemotronh nano omni model;
2. Keep the current rule order and omniRule unchanged.

caihui-lol:
update nemotron rule to append audioTextToText/videoTextToText when it matches nemotron_nano_omni, requesting review again.

beiguo218 (approved):

pallasathena92 (approved):

caihui-lol:
Hi @slin1237 can i get one more review on this PR, thanks so much!

slin1237 (approved):

PR #678, opened 2026-07-21
Add component metadata exclusion support

kunzheng926:
## What this PR does

Adds a generic metadata propagation safeguard for InferenceService components.

This PR introduces `ome.io/component-metadata-exclusions`, a comma-separated annotation on an InferenceService. When present, OME preserves the listed metadata keys on the top-level InferenceService but filters them out when constructing Engine, Decoder, and Router component annotations/labels.

Example:

```yaml
metadata:
  annotations:
    ome.io/component-metadata-exclusions: "example.com/request-id,example.com/workflow-id"
```

## Why we need it

Some metadata is useful for top-level correlation and debugging but should not be propagated into generated component metadata. Dynamic metadata that changes per request or operation can otherwise cause unnecessary rollout churn when copied into component workload metadata.

This keeps the behavior provider-neutral: OME does not need to know which metadata keys are dynamic or vendor-specific. The creator of the InferenceService decides which parent metadata should not flow into generated components.

Fixes #

## How to test

```bash
GOCACHE=/private/tmp/ome-debug-endpoint-dac-oss-ome/.gocache go test ./pkg/controller/v1beta1/inferenceservice/status ./pkg/controller/v1beta1/inferenceservice/components
GOCACHE=/private/tmp/ome-debug-endpoint-dac-oss-ome/.gocache make test
```

Both passed locally.

## Checklist

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -427,11 +427,16 @@ const (
 // DefaultModelLocalMountPath is where models will be mounted by the storage-initializer
 const DefaultModelLocalMountPath = "/mnt/models"
 
+// ComponentMetadataExclusionsAnnotationKey lists comma-separated parent metadata keys that
+// should not be propagated to component metadata.
+const ComponentMetadataExclusionsAnnotationKey = "ome.io/component-metadata-exclusions"
+
 var (
 	ServiceAnnotationDisallowedList = []string{
 		autoscaling.MinScaleAnnotationKey,
 		autoscaling.MaxScaleAnnotationKey,
 		StorageInitializerSourceUriInternalAnnotationKey,
+		ComponentMetadataExclusionsAnnotationKey,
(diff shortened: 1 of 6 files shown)


slin1237 (changes requested):
Pod level event doesn't belong here
If we look at deployment status, which is top level of replicaSet, and replicaSet owns pods
None of them do stuff like this

kunzheng926:
> Pod level event doesn't belong here If we look at deployment status, which is top level of replicaSet, and replicaSet owns pods None of them do stuff like this

Thanks, agreed. I removed the pod-level scheduling/status propagation from this PR.

kunzheng926:
Closing this PR because the debug-capability change is being narrowed to MP-side read-only event collection keyed by the existing endpoint resource, so no open-source OME controller change is needed for this iteration.

PR #726, opened 2026-08-11
[Core] Multi-cluster: IR projection + helpers

fanyang-real:
**Part 1 of 5** — multi-cluster placement control plane, split for reviewability. Merge in order 1→5.

**Series:**
1. #726 — IR projection + helpers *(this PR, foundation)*
2. #727 — WorkloadCluster transport
3. #728 — placement + endpoint publisher
4. #729 — controller config
5. #730 — manager wiring + RBAC

Adds the InferenceReplica projector (projects an ISVC Component into a per-(ISVC, Component) InferenceReplica and reads its authoritative status back) plus the shared helpers it needs:
- `pkg/utils/intstr.go`: `AvailabilityFloor`, the min-available floor helper.
- `inferenceservice/status/top_level_condition.go`: derive a Component's top-level readiness Condition from its `LifecycleStatus`.
- `inferenceservice/reconcilers/irprojector`: the projection + status-read package consumed by the placement admission path (#728).

This is the foundation of the series; it builds standalone.

> Note: the repo is squash-only, so this series is 5 cumulative PRs against `main` (fork branches can't be PR bases). After each merges, the remaining PRs rebase onto the new `main`.

## Summary by CodeRabbit

- **New Features**
  - Added automated management of inference replica resources for supported deployment components.
  - Component settings, labels, annotations, pod configuration, lifecycle, topology, pause state, and autoscaling are now reflected in replica resources.
  - Inference replica status is aggregated into component and top-level service status.
  - Readiness conditions now account for replica availability and disruption-budget settings.

- **Bug Fixes**
  - Avoids unnecessary updates and preserves autoscaler-managed replica counts.
  - Handles missing resources, concurrent creation, deletion, and transient update conflicts more reliably.

### DIFF
--- a/pkg/controller/v1beta1/inferenceservice/reconcilers/irprojector/irstatus_read.go
+++ b/pkg/controller/v1beta1/inferenceservice/reconcilers/irprojector/irstatus_read.go
@@ -1,0 +1,41 @@
+package irprojector
+
+import (
+	"context"
+	"fmt"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/ome/pkg/apis/ome/v1beta1"
+)
+
+// ComponentIRStatus returns the authoritative InferenceReplica status for one
+// Component of an ISVC, read via the supplied reader (cached client is fine —
+// these are non-destructive coordination reads). Returns (nil, nil) when the IR
+// does not exist yet, so callers treat "no authoritative status" the same as the
+// pre-migration empty projection. Any other read failure is returned as a
+// wrapped error — callers MUST distinguish it from the missing-IR case: a
+// transient read error is not "no observation yet", and safety gates that treat
+// it that way fail open. A nil reader is likewise an error, not a missing IR —
+// it signals a wiring bug the caller must not silently absorb. This is the
+// single source of truth every in-cluster decision reads instead of the ISVC's
+// copied LifecycleStatus.
+func ComponentIRStatus(ctx context.Context, reads client.Reader, namespace, isvcName string, c v1beta1.ComponentType) (*v1beta1.InferenceReplicaStatus, error) {
+	// A nil reader is a wiring bug, not an absent IR. Surface it as an
+	// error so safety gates fail closed instead of mistaking it for the
+	// missing-IR case, while still not panicking the reconcile.
+	if reads == nil {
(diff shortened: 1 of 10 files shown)


fanyang-real:
Rebased onto latest `main` and addressed all four review comments in `d3d6c30`:

1. **`irstatus_read.go` — nil reader failed open.** Fixed. `ComponentIRStatus` now returns a wrapped error naming the IR key and component instead of `(nil, nil)`, so safety gates fail closed rather than mistaking a wiring bug for a missing IR. Doc comment updated to state the contract; `TestComponentIRStatus_NilReaderReturnsNil` renamed to `TestComponentIRStatus_NilReaderErrors` and now asserts the error. No production callers existed yet, so tightening the contract here is non-breaking.

2. **`status.go` — detached doc comment.** Fixed. The `aggregateOneComponent` prose was glued to `componentExtFor`'s doc block with no blank line, so godoc attached all of it to `componentExtFor`. Split, with each comment now directly preceding the function it names.

3. **`top_level_condition.go` — zero floor reported Ready at zero serving.** Fixed, and it was a real bug. With `MaxUnavailable: "100%"` (or an integer at/above `Replicas`), `AvailabilityFloor` returns 0, so `ServingReplicas >= floor` evaluated `0 >= 0` and set `Status=True, Reason=MinimumAvailable` for a component serving no traffic — `kubectl wait --for=condition=Ready` would succeed on a fully-down component. The floor is now clamped to at least 1 (`Replicas > 0` is already guaranteed at that point). Clamped only at this call site, not in `AvailabilityFloor` itself, since a zero floor is legitimate for PDB math.

4. **`intstr.go` — clamp bypassed by narrowing.** Fixed. Comparison now happens in `int` before narrowing. Verified this was exploitable: `ScaledCountFromIntOrString("99999999999%", math.MaxInt32, true)` returned **-1021474816** before the fix.

Added regression tests for the two behavioral fixes, both of which I confirmed fail against the unfixed code:
- `pkg/utils/intstr_test.go` — table tests for `ScaledCountFromIntOrString` and `AvailabilityFloor`, including the int32-overflow case. These helpers had no direct coverage.
- `pkg/controller/v1beta1/inferenceservice/status/top_level_condition_test.go` — covers all four condition branches plus both zero-floor paths (percent and integer budgets), and pins that the floor-of-1 does not mask real progress (1 serving Instance under a permissive budget is still Ready).

`gofmt` and `go vet` clean; the affected packages pass. PRs #727–#730 were restacked onto this commit — their own patches are byte-identical, only the base moved.

PR #734, opened 2026-08-11
[Core] Add shared Hugging Face artifact identity and state repository

op109lvb:
## What this PR does

Adds the foundation for shared Hugging Face artifact tracking in model-agent.

This includes HF artifact identity parsing from model annotations, canonical artifact key/path generation, shared parent artifact state in the node ConfigMap, and a repository for lock-safe parent/child reference updates.

## Why we need it

HF-origin models can resolve to the same immutable artifact by `hf-model-id` + `hf-model-sha`. A durable node-local state model lets model-agent track the shared parent artifact and its child model references safely.

This also provides a migration path for the existing direct HF reuse and the foundation for HF-origin OCI Object Storage reuse.

This PR does not change model download behavior yet.

## PR Plan

This is the first PR in the staged model-agent work:

1. Shared HF artifact identity and state repository.
2. Shared artifact lifecycle planning.
3. HF-origin OCI Object Storage artifact reuse.
4. Model agent startup recovery, shared parent validation and incomplete-state hardening.
5. Direct HF source reuse cleanup.

## Changes Made

- Added HF artifact identity and shared artifact ConfigMap entry types.
- Added metadata validation and canonical key/path helpers.
- Added repository operations for artifact lock, status, reference, and delete updates.
- Added `HfArtifactKey` on model entries for child-to-parent lookup.
- Added unit coverage for identity validation and repository state transitions.

## How to test

```bash
go test ./pkg/modelagent ./pkg/constants -count=1
go test -race ./pkg/modelagent -count=1
go vet ./pkg/modelagent
```

## Checklist

- [x] Docs updated (if applicable)

@coderabbitai ignore

## Summary by CodeRabbit

- **New Features**
  - Added support for sharing Hugging Face model artifacts locally across workloads.
  - Added model and revision identity tracking for shared artifacts.
  - Added artifact lifecycle management, including preparation, readiness, failure recovery, locking, and cleanup.
  - Model records now retain references to associated shared artifacts.

- **Reliability**
  - Added safeguards against stale locks, invalid metadata, conflicting references, and accidental removal of in-use artifacts.
  - Improved cleanup of artifacts that are no longer referenced.


pkg/modelagent/configmap_reconciler.go
```diff
@@ -387,6 +391,17 @@ func (c *ConfigMapReconciler) restoreModelInConfigMap(modelID string, cacheEntry
 			if _, exists := cm.Data[modelID]; exists {
 				return
 			}
+			artifactKey, referenceErr := findHfArtifactKeyForModel(cm.Data, modelID)
```
pallasathena92: this block from line 394 to 404 is almost the same code to line 347 to 357?

pkg/modelagent/hf_artifact_repository.go
```diff
@@ -0,0 +1,516 @@
...
+			continue
+		}
+		artifact, err := decodeHfArtifactEntry(key, raw)
+		if err != nil {
+			return "", err
+		}
+		if artifact.Key != key {
+			return "", fmt.Errorf("Hugging Face artifact entry %s records key %s", key, artifact.Key)
+		}
+		if err := validateHfArtifactIdentityAndPath(artifact); err != nil {
```
pallasathena92: **[Design] Self-healing restore is fail-closed on any corrupt artifact entry — including foreign ones.**

This validation runs for *every* `artifact.huggingface.*` entry in the ConfigMap, not just the one referencing `modelKey`. Any single decode/key/path/status failure returns an error, which propagates up to `restoreModelInConfigMap` and aborts the restore of `modelID` entirely.

Since restore is the node's periodic self-healing path, once later PRs start writing artifact entries, one corrupt/partial artifact entry (even for an unrelated model) would block recovery of unrelated model entries node-wide — and the corrupt entry can't be self-healed either.

No live impact in this PR (nothing writes artifact entries yet, so the scan always returns `("", nil)`), which is exactly why it's cheap to decide now. Fail-closed is defensible for durable state, but consider scoping the strict validation to entries whose `Children` actually contain `modelKey` (decode-and-skip foreign entries; keep only the "referenced by multiple artifacts" cross-check fatal) so a foreign corrupt entry can't stall unrelated restores.

pkg/modelagent/hf_artifact_repository.go
```diff
@@ -0,0 +1,516 @@
...
+	if artifact.Key != expectedName {
+		return fmt.Errorf("Hugging Face artifact key %s does not match identity; expected %s", artifact.Key, expectedName)
+	}
+	if strings.TrimSpace(artifact.LocalPath) == "" {
+		return fmt.Errorf("Hugging Face artifact path is empty")
+	}
+	expectedSuffix := filepath.Join(
+		constants.ModelArtifactsDirectory,
+		filepath.FromSlash(artifact.Identity.ModelID),
+		artifact.Identity.CommitSHA,
```
pallasathena92: `expectedSuffix` is built from `artifact.Identity.CommitSHA` verbatim, but both `canonicalHfArtifactPath` and `hfArtifactConfigMapKey` lowercase the SHA. Because `isValidHfCommitSHA` accepts uppercase hex, an entry with an uppercase `Identity.CommitSHA` and a (correctly lowercased) canonical `LocalPath` would fail this check.

It doesn't trigger today because `newHfArtifactIdentity` lowercases on construction, but the three derivations should agree. Suggest `strings.ToLower(artifact.Identity.CommitSHA)` here (and/or make `newHfArtifactIdentity` the sole constructor of identities).

pkg/modelagent/hf_artifact_repository.go
```diff
@@ -0,0 +1,516 @@
...
+		}
+		if strings.TrimSpace(modelPath) == "" {
+			return false, fmt.Errorf("Hugging Face artifact %s has an empty path for model %s", stored.Key, modelKey)
+		}
+
+		delete(stored.Children, modelKey)
+		model.HfArtifactKey = ""
+		result = RemoveModelReferenceResult{Artifact: stored, ReferenceRemoved: true}
+		if len(stored.Children) == 0 {
+			stored.Status = HfArtifactStatusUpdating
```
pallasathena92: Dropping the last reference flips the artifact to `Updating` under a fresh deletion `LockID`, expecting the caller to delete files then call `DeleteIfUnreferenced`. If the process dies in between, the entry is stuck: `TryAcquireLock*` refuses to re-acquire an `Updating` entry, and `DeleteIfUnreferenced` requires the now-lost lock.

pkg/modelagent/configmap_reconciler.go
pallasathena92: **[Nit] Duplicated recover-and-marshal block.**

This `findHfArtifactKeyForModel` → set `HfArtifactKey` → `json.Marshal` → `cm.Data[modelID] = …` sequence duplicates the initial-write path above (~L347), differing only in `err` vs `referenceErr`. That's the kind of near-duplicate that drifts over time — extracting a small closure would remove the divergence risk.

pallasathena92 (commented):
A few review notes on the shared-artifact state foundation, posted inline. None block merge; #1 and #2 are the ones worth acting on before later PRs build on this.

pkg/modelagent/hf_artifact_repository.go
op109lvb: Good point. This behavior belongs to the planned recovery and hardening PR. This PR is limited to the shared-artifact state and repository foundation, so I removed the restore logic from this PR and will re-introduce it with the complete recovery design.

pkg/modelagent/hf_artifact_repository.go
op109lvb: Addressed

pkg/modelagent/configmap_reconciler.go
op109lvb: Agreed. This duplication is part of the same ConfigMap recovery path. I removed it with the recovery logic from this foundation PR. The planned recovery PR will re-introduce a single consolidated restore path.

pkg/modelagent/hf_artifact_repository.go
```diff
@@ -0,0 +1,480 @@
...
+			stored.LockID = uuid.NewString()
+			stored.LastCompletedLockID = ""
+			result.Artifact = stored
+			result.LastReferenceRemoved = true
+		}
```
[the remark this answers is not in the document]
op109lvb: The recovery design needs lock fencing around filesystem deletion, not only an expiry timestamp. This will be introduced in the following recovery PR.

PR #773, opened 2026-08-19
[Core] Add the workload types package

fanyang-real:
First slice of the InferenceReplica runtime, which is the piece that makes `OMENative` actually render pods. Authored by @slin1237; I am porting it.

## What this is

The vocabulary the rest of the runtime is written in: the per-Instance plan and operation kinds, Instance and Runner status shapes that mirror the CRD field-for-field, retry-block and expectations bookkeeping, migration and termination records, event reasons, condition helpers, and the dependency and service seams the reconciler is constructed with.

## Why it is first

It imports **no OME packages at all** — only the standard library, k8s API machinery and controller-runtime — so it stands alone, builds and tests on its own, and every later slice can be reviewed against a type vocabulary already in the tree.

Nothing in the tree consumes it yet. That is deliberate: the alternative is one very large PR.

## The wave this belongs to

The dependency graph is a clean DAG, so the remaining slices land in topological order:

```
workload/types            <- this PR   (16 files)
workload/podreadiness, workload/drain  (4 files)   no in-tree deps
workload/audit, workload/query         (15 files)
workload/revision, workload/podgroup   (4 files)
workload/gang                          (4 files)
workload/ops                           (47 files)  will be split further
workload                               (33 files)
inferencereplica                       (27 files)
```

Together those are what turns an `InferenceReplica` into running pods, plus PodGroup creation and gang admission — the gaps documented in #770.

## Verified

`go build`, `go vet`, `gofmt` and `go test` pass on the package standing alone (57.1% statement coverage from the ported tests). Scanned for upstreaming-boundary problems: no internal identifiers, hostnames, paths, ticket or diff references. One comment was rewritten from diff-narrative ("no longer stamped") into the invariant it was describing, per the repository comment guidance.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive workload lifecycle tracking for updates, restarts, migrations, retries, and pod termination.
  * Added migration phase management, manual and automatic migration support, and gang-scheduling status visibility.
  * Added configurable retry backoff and failure handling for unsuccessful updates.
  * Added workload planning for topology, scaling, surge capacity, runner configuration, and component services.
  * Added clearer pod termination diagnostics and concise failure details.
* **Bug Fixes**
  * Improved handling of pending pod changes through expiration safeguards.
* **Tests**
  * Added coverage for retries, migrations, expectations, scheduling, and termination scenarios.

### DIFF
--- a/pkg/controller/v1beta1/workload/types/conditions.go
+++ b/pkg/controller/v1beta1/workload/types/conditions.go
@@ -1,0 +1,33 @@
+package types
+
+// ConditionType / ConditionReason are workload-internal identifiers
+// stamped on metav1.Condition entries. Values match the legacy
+// omenative/status strings byte-for-byte so operator dashboards keep
+// matching.
+type ConditionType string
+
+func (t ConditionType) String() string { return string(t) }
+
+type ConditionReason string
+
+func (r ConditionReason) String() string { return string(r) }
+
+const (
+	// ConditionGangSchedulingUnavailable is True when the Component
+	// has at least one multi-pod Instance but the scheduler-plugins
+	// PodGroup CRD is missing. The reconciler still creates pods —
+	// gang scheduling is a soft requirement so workloads proceed
+	// without blocking — but partial-gang placement is possible and
+	// the runtime may hang.
+	ConditionGangSchedulingUnavailable ConditionType = "GangSchedulingUnavailable"
+)
+
+const (
+	// ReasonPodGroupCRDNotInstalled stamps the
+	// GangSchedulingUnavailable condition when the
+	// scheduler-plugins PodGroup CRD is missing.
+	ReasonPodGroupCRDNotInstalled ConditionReason = "PodGroupCRDNotInstalled"
+	// ReasonGangSchedulingAvailable stamps Status=False (CRD present,
+	// or Component is single-pod).
+	ReasonGangSchedulingAvailable ConditionReason = "GangSchedulingAvailable"
+)
--- a/pkg/controller/v1beta1/workload/types/deps.go
+++ b/pkg/controller/v1beta1/workload/types/deps.go
@@ -1,0 +1,117 @@
+package types
+
+import (
+	"context"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/clock"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ObservationReader is a watch-backed (cached) reader; it may lag the
+// API server. Safe wherever a later live read or optimistic-concurrency
+// write re-validates the observation.
+type ObservationReader = client.Reader
+
+// AuthoritativeReader reads live from the API server. Required where
+// cache lag is a correctness bug: revision bookkeeping, audit ledger,
+// EndpointSlice drain checks, gang topology proofs.
+type AuthoritativeReader = client.Reader
+
+// Deps is the manager-scoped wiring the workload reconciler needs.
+// Shared across every workload the same controller manages (unlike
+// the per-workload ReconcileInput).
+type Deps struct {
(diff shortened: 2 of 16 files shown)


fanyang-real:
Addressed, and rebased onto latest `main` (it had drifted 7 commits behind). Amended into the single commit so the slice stays one reviewable unit.

**Major — restrict exit-zero termination selection to failed Pods.** Fixed, and thank you — this one was real and reachable. Step 3's own doc comment said *"exit 0 but the pod is Failed"*, but the code had no phase guard, and `allStatuses` appends `InitContainerStatuses`. Every healthy pod with a completed init container — exit 0, `Reason: Completed` — matched step 3 and produced an `InstanceTermination`. It bites through the not-yet-`Failed` scan a restart path uses to attribute a cause, where a healthy pod could supply a bogus `LastFailure`. Step 3 is now guarded on `PodFailed`, so the `restartPolicy=Never` sidecar case it exists for still works. Two tests: one that a cleanly-completed init container yields nil, one that exit 0 is still captured once the pod has Failed.

**Major — apply the explicit non-empty reason override.** Declining; I believe this one is incorrect. The current precedence is the documented contract ("When PodTermination DID extract a record but with an empty Reason, the override fills it in"), and it is already pinned by an existing test asserting that an `ImagePullBackOff` override must *not* clobber an extracted `OOMKilled` exit-137 record. That is the right precedence: the override is a floor for callers that have classified a wedge from live waiting state, not a ceiling over a richer per-container extraction. Making the explicit reason win would discard the exit code with it.

**Minor — expire expectations at the exact deadline.** Fixed: `!Now().Before(Deadline)`. The boundary case is now asserted directly rather than a second past it, and the "past the TTL" assertion moved to its own test so both are covered.

**Minor — handle unknown migration phases explicitly.** Fixed. The terminal phases are now enumerated at rank 4 and the `default` returns -1. Ranking an unrecognized phase above the whole chain was the harmful direction: `MigrationPhaseAtOrPast` would report every advancement as already done and wedge the record permanently. At the bottom, the executor can still drive it forward. Added a full-chain ordering test plus one pinning that `""` and an unknown value are not at-or-past anything.

**Minor — the two `LATENT BUG` doc blocks contradict themselves.** Fixed, and you were right that a reader could not tell what "not honored" meant. Both are rewritten to state the invariant: the field *is* projected by the adapter, nothing in the rollout engine reads it, and the consequence is that the spec field it carries has no effect. The stale "nothing sets this field today" clause is gone. I also dropped the pointers to `inferencereplica/convert.go` and `inferencereplica/reconciler.go` — those files are later slices in this wave and are not in the tree yet, so the references would have dangled for a reader here.

**Minor — correct the nil-return contract.** Fixed: `PodTermination` returns nil for a nil pod *and* for a pod carrying no failure signal.

Verified: `gofmt`, `go build`, `go vet`, `go test` clean. Coverage 57.1% → 60.9%. `go.mod` / `go.sum` untouched. Each of the three behaviour fixes has a test that fails without it.

One housekeeping change while amending: added the DCO `Signed-off-by` trailer this commit was missing, and a `Co-authored-by` for @slin1237 who authored the package — squash-merge rewrites the commit author to the PR opener, so the trailer is what preserves the attribution.

fanyang-real:
Fixed — and this one was mine, introduced by the previous round rather than pre-existing. Ranking an unknown phase at -1 closed the hole on the `p` side but opened the mirror of it on the `target` side: `MigrationPhaseAtOrPast(Accepted, "Bogus")` became true.

`migrationPhaseRank` now returns `(int, bool)` and `MigrationPhaseAtOrPast` reports false unless both phases are in the chain. That keeps the property the first fix was after — an unrecognized phase never reads as "already past", in either position — and false remains the recoverable answer, since it lets the executor drive the record forward instead of wedging it. The regression test now covers the unknown value as `p`, as `target`, and on both sides at once.

Coverage 61.6%. `gofmt`, `go build`, `go vet`, `go test` clean.

PR #776, opened 2026-08-21
[Core] Parallelize OCI object storage model verification (MD5)

fredericsun:
## What this PR does

- Parallelizes OCI model file integrity checks across model files.
- Adds configurable `--model-verification-concurrency`.
- Enforces one pod-wide concurrency limit shared by all model tasks.
- Keeps the default at `1`, preserving serial verification for a single model.
- Preserves existing MD5/size mismatch handling and verification metrics.
- Adds tests for shared concurrency limits and validation failures.

## Why we need it

Final MD5 verification is a significant bottleneck for large, sharded models because files were previously validated sequentially.

Parallel verification significantly reduces this phase. The pod-wide limiter prevents simultaneous model tasks from multiplying CPU and NVMe read pressure beyond the configured limit.

With the default value of `1`, concurrent model tasks also share one verification slot. Deployments must explicitly configure a higher value, such as `8`, to enable the optimization.

## Validation

Validated with `gpt-oss-120b` (195.8 GB) and a pod-wide verification concurrency of 8. Across 72 downloads, single-model final verification decreased from the serial baseline of ~324–329s to ~113–117s, while two concurrent models averaged ~125–127s per model. All 72 MD5 verifications passed, with no checksum mismatch, model-agent restart, or probe failure observed

## How to test

```bash
go test ./pkg/modelagent ./cmd/model-agent
helm lint charts/ome-resources
```
The test verify that:
- The default concurrency is `1`.
- Invalid concurrency values fall back to `1`.
- Multiple model tasks share the same pod-wide limiter.
- MD5/size mismatches and validation errors are still reported.
- The configured concurrency limit is never exceeded.

## Checklist

- [X] Tests added/updated (if applicable)

## Summary by CodeRabbit

* **New Features**
  * Added configurable concurrency for model-file integrity verification.
  * Administrators can set the verification limit through the model agent configuration or deployment settings.
  * Verification defaults to one concurrent check, with invalid values safely normalized to the minimum.
  * Shared concurrency limits now apply across simultaneous model verifications while preserving validation errors and monitoring metrics.


op109lvb (approved):

op109lvb:
OME-agent also performs per-file MD5 validation during replication job. Do you plan to include this in a separate PR?

fredericsun:
> OME-agent also performs per-file MD5 validation during replication job. Do you plan to include this in a separate PR?

Thanks for bringing it up. Could you clarify whether you mean adding a dedicated checksum concurrency limit to OME-agent? OCI-to-OCI already computes MD5 concurrently via `num_connections`, while HF/PVC-to-OCI computes it serially. Are you suggesting we standardize this in a separate PR?

op109lvb:
> > OME-agent also performs per-file MD5 validation during replication job. Do you plan to include this in a separate PR?
> 
> Thanks for bringing it up. Could you clarify whether you mean adding a dedicated checksum concurrency limit to OME-agent? OCI-to-OCI already computes MD5 concurrently via `num_connections`, while HF/PVC-to-OCI computes it serially. Are you suggesting we standardize this in a separate PR?

Both the Hugging Face → OCI and OCI → OCI replication paths need parallel per-file checksum generation and validation.

fredericsun:
> > > OME-agent also performs per-file MD5 validation during replication job. Do you plan to include this in a separate PR?
> > 
> > 
> > Thanks for bringing it up. Could you clarify whether you mean adding a dedicated checksum concurrency limit to OME-agent? OCI-to-OCI already computes MD5 concurrently via `num_connections`, while HF/PVC-to-OCI computes it serially. Are you suggesting we standardize this in a separate PR?
> 
> Both the Hugging Face → OCI and OCI → OCI replication paths need parallel per-file checksum generation and validation.

I will implement HF/PVC-to-OCI optimization in a separate PR. The design will be different from MD5 verification parallelization in model-agent.

truddy0 (approved):

pallasathena92 (commented):
The split of responsibilities here reads well: the per-model worker pool is what lets a single model draw more than one slot, and the shared `verificationLimiter` is what stops the pod from over-reading NVMe when several tasks land at once. Neither alone gets you both properties, so having both is right. A few things I would like to see addressed before this merges.

### The default of `1` is a regression for nodes verifying more than one model

Today verification runs on the download worker goroutine (`runWorker` -> `processTask` -> `downloadModel` -> `verifyDownloadedFiles`), so concurrent model tasks each verify their own files serially and independently. The shipped chart passes `--num-download-worker 2`, and the flag default in `cmd/model-agent/main.go` is `5` — so on `main` a pod can already have 2 (or up to 5) MD5 computations in flight.

After this change, a pod running the default `modelVerificationConcurrency: 1` has exactly one permit for the whole process, so those 2 concurrent models now serialize against each other. Single-model verification is unchanged, but the two-model case gets slower than it is today, and nothing in the flag help or `values.yaml` signals that.

The PR body notes that concurrent tasks share the slot, but I do not think a silent slowdown should be what you get by not setting the new flag. Options, roughly in order of preference:

1. Default the limiter to `numDownloadWorker` so the out-of-the-box behavior matches current throughput, and let operators raise it to 8 for the optimization.
2. Keep `1` as the per-model default but size the pod-wide limiter separately (see below).
3. At minimum, set the chart value to something >= the `--num-download-worker` it ships with, and say in the flag help that this caps the whole pod, not one model.

### `N` is doing two different jobs

`WithModelVerificationConcurrency(n)` sets both the per-model pool size (`min(n, len(uris))` in `verifyDownloadedFilesWithValidator`) and the pod-wide permit count, from one knob. Those are separate concerns: "how parallel may one model be" versus "how much disk pressure may this pod generate in total." As written you cannot express "each model may use up to 4, but the pod up to 12."

It works today because the limiter is always the binding constraint, but the coupling is worth either splitting into two values or calling out in a comment on the option so the next person does not have to derive it.

### `modelVerificationConcurrency` duplicates `limiter.limit()`

The `Gopher` now carries both the `int` and the `*verificationLimiter`, kept in sync only by the option function. `effectiveModelVerificationConcurrency()` re-clamps the `int` even though `newVerificationLimiter` already clamps. Deriving the worker count from `s.modelVerificationLimiter.limit()` and dropping the field removes the second source of truth and the extra clamp. The nil-limiter fallback in `verifyDownloadedFilesWithValidator` would need a small rework, but that path only exists for zero-value `Gopher{}` in tests — worth confirming it is not reachable in production and, if so, replacing it with the limiter always being non-nil from `NewGopher`.

### Nit: the job is claimed before the permit is acquired

Each worker receives from `jobs` and only then calls `limiter.acquire(ctx)`, so a blocked worker sits on a claimed file. Harmless here since every file is equivalent work and `jobs` is unbuffered, but acquiring before the receive would make the flow easier to reason about.

### Nice to have

`IsLocalCopyValid` does a `HeadObject` round trip per file before reading the file, so for models with many small files this is also hiding network latency, not just spreading disk reads. That is a real part of the win and might be worth a line in the PR description alongside the gpt-oss-120b numbers.

fredericsun:
Run a new round of validation with Kimi K3 (1.56 TB, 118 files), using the same GPU node and resources (10 CPU / 100 GiB):
- Final MD5 verification: **44m 4s → 6m 28s** at concurrency 8 — **85.3% less time, 6.81× faster**.
- Download + verification: **61m 59s → 23m 29s**.
- All files passed verification; model reached Ready with **zero model-agent restarts**.

pallasathena92 (approved):

XinyueZhang369 (approved):

Issue #778, opened 2026-08-21
BenchmarkJob cannot schedule against PVC-backed models; --api-model-name hardcoded to "vllm-model"


weetime:
## Summary

`BenchmarkJob` still assumes every `BaseModel` is HostPath-backed and downloaded per node. That assumption was fixed for the serving path in #656 ("Mount PVC-backed models into ISVC pods"), whose description explicitly calls itself *"the serving-side half of PVC support"* — the benchmark controller never got the other half.

As a result, a `BenchmarkJob` targeting an InferenceService whose model is `pvc://` can **never** be scheduled. There is also a hardcoded model name that makes every benchmark request 404 against a normally-configured runtime.

Verified against `main` @ 86184c3.

## 1. PVC-backed models make the benchmark pod permanently unschedulable

`pkg/controller/v1beta1/benchmark/controller.go:250` adds the model-ready node selector unconditionally:

```go
isvcutils.AddNodeSelectorForModelReadyNode(podSpec, baseModelMeta)
```

and `controller.go:354` builds the model volume as HostPath only:

```go
HostPath: &v1.HostPathVolumeSource{Path: *baseModel.Storage.Path},
```

Both assume `model-agent` has staged the weights onto the node. But `model-agent` **skips PVC storage by design** — `pkg/modelagent/gopher.go`:

```go
case storage.StorageTypePVC:
    s.logger.Infof("Skipping PVC storage type for model %s (handled by BaseModel controller)", modelInfo)
    return nil          // returns before the node-labeling step
```

and `pkg/modelagent/scout.go` returns `false` from `shouldDownloadModel` for PVC as well. So **no node ever carries `models.ome.io/<ns>.basemodel.<name>=Ready` for a PVC-backed model**, and the benchmark pod stays Pending forever:

```
0/2 nodes are available: 2 node(s) didn't match Pod's node affinity/selector.
preemption: 0/2 nodes are available: 2 Preemption is not helpful for scheduling.
```

This is precisely the failure mode #656 anticipated and guarded against on the serving side (`components/base.go`, `UpdatePodSpecNodeSelector`):

```go
// Skip the per-node model-ready affinity for PVC-backed models. PVCs
// aren't tied to specific nodes and the model agent never labels nodes
// for them, so the readiness selector would leave the pod unschedulable.
if isPVCBaseModel(b) { ... return }
```

The benchmark controller has no equivalent branch.

[...]

PR #779, opened 2026-08-21
[Bugfix] benchmark: derive --api-model-name

weetime:
Fixes the second problem reported in #778.

## What this fixes

`BuildInferenceServiceArgs` passed the literal string `"vllm-model"` as `--api-model-name`:

```go
args := map[string]string{
    "--api-key":         "sample-key", // TODO: Use actual service account key later
    "--api-model-name":  "vllm-model",
    "--model-tokenizer": *baseModel.Storage.Path,
}
```

That value becomes the `model` field of every OpenAI request genai-bench sends. vLLM and SGLang validate it against their `--served-model-name` and return 404 on a mismatch, so unless a runtime happens to serve under the literal name `vllm-model`, every request in the benchmark fails.

The failure is quiet, which is what makes it costly: the BenchmarkJob reports `Completed`, the pod exits 0, and a full set of result files and plots is written. Only `num_error_requests` inside the per-run JSON reveals that nothing succeeded.

Note the neighbouring `--api-key` line carries a `// TODO`; this one did not, so it read as an intentional constant rather than a placeholder.

## Approach

The correct value is already in scope — the function holds both the `InferenceService` and the `BaseModel`. Preference order:

1. the first `--served-model-name` value on the InferenceService's engine runner, when the user overrode the runtime's command there;
2. otherwise the `BaseModel` name, which is how OME identifies a model elsewhere.

The flag is variadic in both engines, so only the first value is taken — later entries are aliases. A dangling flag (nothing after it, or another flag) falls through to the BaseModel name rather than consuming an unrelated token.

Not covered: a runtime that sets `--served-model-name` to something unrelated to the model name and is not overridden by the InferenceService. Resolving that would mean fetching the ServingRuntime here; it seemed out of scope for a bug fix, but I'm happy to extend it if you'd prefer.

`--api-key` is left as `"sample-key"` with the `TODO` replaced by a note explaining why a placeholder is acceptable there (genai-bench requires the flag to be non-empty for the openai backend but does not verify it).

## Tests

`TestResolveAPIModelName` covers seven cases: fallback to the BaseModel name, nil InferenceService, resolution from the runner command, resolution from the runner args, only-the-first-name-of-a-variadic-flag, a dangling flag followed by another flag, and the flag at the very end of the token list.

The test fails without the fix — with the old constant in place, four of the seven cases report `resolveAPIModelName() = "vllm-model"`.

## Verified on a cluster

Kubernetes 1.30, OME v1.2.2-based build with this patch applied on top, vLLM 0.8.5.post1 on a V100, BaseModel `qwen2-5-0-5b-instruct`.

The runtime serves under its own name only — no `vllm-model` alias — which is exactly the case that used to 404:

```
$ curl .../v1/models
served: ['qwen2-5-0-5b-instruct']
```

The generated benchmark pod now carries the derived name:

```
--api-model-name qwen2-5-0-5b-instruct
```

and the run completes clean:

```
conc=1  done=17  err=0  rps=5.37  tput=868.9   ttft_p95=0.1361
conc=2  done=18  err=0  rps=8.98  tput=1457.7  ttft_p95=0.0626
```

Before the fix, the same setup required adding `vllm-model` as a second `--served-model-name` on the runtime to get any successful request at all.

## Scope

This PR deliberately covers only the `--api-model-name` problem. The PVC-related half of #778 (node selector and model volume) is a separate change and I'd rather send it once I can verify it on a cluster the same way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved benchmark compatibility with custom inference service configurations by using the configured served model name when available.
  * Added fallback behavior to use the base model name when no served model name is configured.
  * Ensured benchmark API requests include a valid API key value as required by the benchmarking tool.

* **Tests**
  * Added coverage for custom model names, fallback behavior, repeated flags, and invalid or missing values.

### DIFF
--- a/pkg/controller/v1beta1/benchmark/utils/utils.go
+++ b/pkg/controller/v1beta1/benchmark/utils/utils.go
@@ -3,6 +3,7 @@ package benchmarkutils
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -67,8 +68,11 @@ func BuildInferenceServiceArgs(ctx context.Context, c client.Client, endpointSpe
 		}
 
 		args := map[string]string{
-			"--api-key":         "sample-key", // TODO: Use actual service account key later
-			"--api-model-name":  "vllm-model",
+			// genai-bench requires a non-empty api-key for the openai backend but
+			// does not verify it; engines fronted by OME don't authenticate either.
+			// A real key would have to come from the spec or a Secret.
+			"--api-key":         "sample-key",
+			"--api-model-name":  resolveAPIModelName(inferenceService, baseModelName),
 			"--model-tokenizer": *baseModel.Storage.Path,
 		}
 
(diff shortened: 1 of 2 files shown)


weetime:
Marking this draft — my impact analysis above was incomplete in a way that matters, and as written this change would **regress** the bundled vLLM runtimes. Correcting the record rather than leaving it to be found in review.

## What I got wrong

**1. The bundled runtimes are not affected by the hardcoded value.**

I counted them: all 17 runtime definitions under `config/runtimes/vllm/` set `--served-model-name` to the literal `vllm-model`. So the hardcoded `"vllm-model"` matches them exactly. It is a convention the benchmark controller is aligned with, not an oversight.

**2. SGLang does not validate the model field at all.**

`python/sglang/srt/openai_api/adapter.py` (v0.4.6.post4) echoes the request's model back without comparing it to the served name — no 404 path. So the 189 runtime definitions under `config/runtimes/srt/`, which use HuggingFace ids, are unaffected too.

Net: **none of the bundled runtimes are affected.** My original phrasing ("every request in the run fails") was true only in the abstract and omitted this, which made the problem look far larger than it is.

**3. This patch would regress the bundled vLLM runtimes.**

With no runner override on the InferenceService — the normal case — my code falls back to the BaseModel name. Against a bundled vLLM runtime serving as `vllm-model`, that turns a working benchmark into 404 on every request. That is a regression introduced by this PR, not a fix.

## What still holds

vLLM does return 404 on a mismatch. Verified on a cluster, against a service serving only under its own name:

```
qwen2-5-0-5b-instruct → 200
vllm-model            → 404 {"message":"The model `vllm-model` does not exist."}
```

So the real, narrower gap is: **a user-defined vLLM runtime whose `--served-model-name` is not `vllm-model` cannot be benchmarked.** The failure is silent — the BenchmarkJob reports `Completed`, the pod exits 0, all result files and plots are written, and only `num_error_requests` in the per-run JSON shows nothing succeeded.

Whether that is a defect or simply an undocumented convention ("name your vLLM runtime `vllm-model` if you want to benchmark it") is a maintainer call, which is why I'd rather ask than push a patch.

## Proposed direction

Keep `vllm-model` as the default so bundled runtimes are untouched, and add an optional escape hatch — an optional `modelName` on the endpoint spec, mirroring `Endpoint.ModelName`, which the `inferenceService` branch has no equivalent of today. Unset means current behaviour; set means send that instead. No new resource lookups, no behaviour change for anything that works today.

That is an API addition, so I've asked in #778 whether it needs an OEP and whether the shape is acceptable before implementing it. Happy to close this in favour of a docs-only change if you consider the convention sufficient.

weetime:
Closing this. The premise was wrong.

I filed it believing a BenchmarkJob had no way to specify the model name the client sends. It does: `spec.endpoint.endpoint.modelName` is passed straight through as `--api-model-name`. The `"vllm-model"` constant in the `inferenceService` branch is the default for the case where the user supplied no name, not a hole in the API — and it matches the `--served-model-name` used by every runtime under `config/runtimes/vllm/`.

I should have read `BuildInferenceServiceArgs` end to end before opening anything. Sorry for the noise.

PR #782, opened 2026-08-21
[Core] Add workload readiness and drain packages

fanyang-real:
## What this is

Two pure leaf packages that the operation packages are built on.

**`workload/podreadiness`** — the multi-writer readiness gate protocol for the controller-owned `ome.io/serving` pod condition. OMENative has several overlapping reasons to hold a pod NotReady at the same time (migration source drain, in-place update drain, restart drain, scale-down drain). A single binary gate has those writers overwrite each other, so the condition's `Message` field carries a JSON list of holders instead, and a pod is Ready only once the list is empty and kubelet reports containers ready. Writes go through `Status().Patch` with strategic merge on the condition list, and every patch pins the `resourceVersion` it was computed against — a stale base 409s and retries rather than silently dropping another writer's hold.

**`workload/drain`** — the EndpointSlice-convergence signals that gate destructive pod operations: has this pod left rotation and is safe to delete, and is this surge pod eligible for traffic. It distinguishes "no Service at all" (drain trivially complete) from "Service exists but slices have not propagated yet" (cold cache, kube-controller-manager lag), so a cold read cannot report a drain complete while the pod is still serving.

## Why these are next

Neither package imports anything from OME — only the standard library, k8s API machinery and controller-runtime. They stand alone, build and test on their own, and add no module dependencies (`go.mod`/`go.sum` unchanged).

Nothing in the tree consumes them yet. That is deliberate; the alternative is one very large PR.

## The wave this belongs to

```
workload/types                         <- #773
workload/podreadiness, workload/drain  <- this PR
workload/audit, workload/query
workload/revision, workload/podgroup
workload/gang
workload/ops                           will be split further
workload
inferencereplica
```

## Verified

`gofmt`, `go build`, `go vet` and `go test` pass on both packages standing alone — `drain` 95.1% statement coverage, `podreadiness` 83.0%, from the ported tests. `go.mod` and `go.sum` are untouched.

Scanned for upstreaming-boundary problems: no internal identifiers, hostnames, paths, ticket or diff references. Two comments were rewritten from diff/roadmap narrative ("we'd previously fail-open", "v1 of the multi-pod path does NOT…") into the invariants they were describing, per the repository comment guidance.

## New Features

- Added endpoint-based pod drain and rotation checks for safer traffic transitions.
- Added efficient batched evaluation for multiple pods associated with a service.
- Added multi-writer serving readiness support, enabling independent readiness holds and lifecycle updates.
- Added readiness queries for serving, container, and overall pod readiness.

## Bug Fixes

- Improved handling of terminating, unavailable, duplicate, and not-yet-materialized endpoints.
- Added validation and recovery for stale, malformed, or inconsistent readiness state.
- Preserved concurrent readiness updates during conflict-safe changes.
- Improved handling of missing services, pod identity changes, and readiness update conflicts.

### DIFF
--- a/pkg/controller/v1beta1/workload/drain/drain.go
+++ b/pkg/controller/v1beta1/workload/drain/drain.go
@@ -1,0 +1,332 @@
+// Package drain owns the EndpointSlice-convergence signals OMENative
+// uses to gate destructive pod operations: IsPodDrained (safe to
+// delete?) and IsPodInRotation (surge eligible for new traffic?).
+//
+// Boundary: pure leaf. Only Kubernetes API machinery. Knows nothing
+// about ReconcileParams or any OMENative-specific shape — every entry
+// point takes a plain client.Reader + (namespace, serviceName, pod).
+package drain
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// IsPodDrained reports whether pod is no longer routable through serviceName.
+// It returns true when, across every EndpointSlice for serviceName in the
+// pod's namespace, either:
+//   - no endpoint targets pod, or
+//   - the endpoint that targets pod has Conditions.Ready != true.
+//
+// The returned bool is the EndpointSlice-convergence signal used during
+// drain: once true, kube-proxy will have removed (or marked not-ready)
+// the pod's address and new connections to serviceName stop landing on
+// this pod. Existing connections and runtime-level shutdown are out of
+// scope — the caller layers gracePeriodSeconds on top of this signal.
+//
+// reader is taken as client.Reader (not client.Client) so callers may
+// pass a live API reader to bypass the cached controller-runtime read
+// when cache lag would make a drain falsely declare done. The cached
+// client works as well and is what tests use.
+func IsPodDrained(ctx context.Context, reader client.Reader, namespace, serviceName string, pod *corev1.Pod) (bool, error) {
+	if pod == nil {
+		return false, fmt.Errorf("IsPodDrained: nil pod")
+	}
+	if serviceName == "" {
+		return false, fmt.Errorf("IsPodDrained: empty serviceName")
+	}
+
+	slices, err := EndpointSlicesForService(ctx, reader, namespace, serviceName)
+	if err != nil {
+		return false, err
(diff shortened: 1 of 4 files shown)


fanyang-real:
Addressed. Pushed as an amend to the single commit so the port stays one reviewable slice.

**Major — Pod identity check on the remove path.** Fixed. `removeNotReadyKey` now compares `fresh.UID` against `pod.UID` right after the re-read: `ErrPodIdentityChanged` on the strict path, and `nil` on the `ignoreNotFound` variant, where a same-name replacement is the same no-op as a vanished pod. Good catch on the `MarkPodServingWithChange` consequence — a `changed=true` there would drain the predecessor on a promotion the replacement never earned, so there is a regression test pinning that specifically. Four tests added; three fail without the fix, and the fourth pins that an observation carrying no UID still releases (callers that synthesize a pod reference from a slot name have nothing to compare against).

**`LastTransitionTime` when `Status` does not change.** Fixed at both patch sites via a small `transitionTime` helper. Writers join and leave the message list constantly while `Status` stays `False`, so restamping on each one reset the "how long has this pod been NotReady" clock. The helper deliberately only preserves the timestamp when the stored status already equals the one being written, so the Add path's self-heal branch — `True` with a non-empty list, flipping back to `False` — still stamps a fresh transition. Test covers hold → join → partial release → final release.

**`IsPodInRotation` no-slice paths.** Added: no slices at all, slices that target another pod, LIST error propagation through `countingReader`, and nil-pod / empty-serviceName rejection.

**Namespace/kind matrix through the package-level path.** Added, as a parity assertion rather than a duplicated expectation table — the direct call is asserted equal to the `Batcher` result, so the test fails if either implementation drifts alone.

**Test `Get` error and condition pointer checks.** Fixed in `TestMultiWriter_TwoKeysHoldNotReady` via local `reread` / `condition` helpers, so a failed read or a missing condition fails the test instead of nil-panicking.

**Declining: one shared `TargetRef` matcher for both paths.** Reasonable in isolation, but this file is a port from the repo it originated in, and a structural refactor here would diverge the two trees for no behaviour change — it makes every later slice in this wave harder to review. The new parity assertion covers the actual risk (the two implementations drifting apart) without the divergence. Happy to do the consolidation upstream-first as a follow-up once the wave has landed.

On the docstring-coverage warning: the 84 functions counted are mostly table-driven test closures and single-purpose test functions whose names carry their intent. The exported API in both packages is documented.

Verified: `gofmt`, `go vet`, `go test` clean. Coverage is up — `drain` 95.1% → 99.0%, `podreadiness` 83.0% → 83.8%. `go.mod` / `go.sum` still untouched.

PR #786, opened 2026-08-24
[Bugfix] benchmark: read served model name from engine pods

### LINKED ISSUE #781: [BUG] BenchmarkJob sends a hardcoded model name; requests 404 while the job reports Completed
## What happened?

`BenchmarkJob` sends a hardcoded string as the OpenAI `model` field. Every request 404s, but the job still reports `Completed`.

`BuildInferenceServiceArgs` (`pkg/controller/v1beta1/benchmark/utils/utils.go:71`, `main` @ 23c97f19) builds the genai-bench arguments for the `endpoint.inferenceService` path:

```go
args := map[string]string{
    "--api-key":         "sample-key", // TODO: Use actual service account key later
    "--api-model-name":  "vllm-model",
    "--model-tokenizer": *baseModel.Storage.Path,
}
```

`--api-model-name` becomes the `model` field of every request genai-bench sends. vLLM validates it against `--served-model-name` and returns 404 on a mismatch, so a runtime that does not happen to serve under the literal name `vllm-model` fails on every single request.

**The failure is silent, which is what makes it expensive.** The pod exits 0, the BenchmarkJob status goes to `Completed`, and a complete set of result files, spreadsheets and plots is written to the output location. Only `num_error_requests` inside the per-run JSON reveals that nothing succeeded. A user reading the dashboard has no signal that the numbers are meaningless.

### Scope

This is narrower than it first looks, and the scoping is worth stating so it can be weighed properly:

| | count | `--served-model-name` |
|---|---|---|
| `config/runtimes/vllm/` | 17 | all `vllm-model` |
| `config/runtimes/srt/` | 189 | 167 distinct HuggingFace ids, none `vllm-model` |

SGLang echoes `request.model` back without comparing it to the served name, so the SGLang half has no 404 path. **None of the bundled runtime definitions are affected** — the hardcoded value matches the vLLM catalogue's convention, and SGLang's leniency covers the rest. That is why it has held up.

What breaks is a **user-defined vLLM runtime** whose `--served-model-name` is anything else, which is the normal case once you bring your own runtime.

Background: this was one of two problems reported in #778. That issue bundled it with a separate PVC-scheduling problem and has been closed; this one is filed on its own so it can be judged on its own.

## What did you expect to happen?

Either the benchmark uses the name the engine actually serves under, or it fails loudly instead of producing a full set of plausible-looking results from 100% failed requests.

## How can we reproduce it (as minimally and precisely as possible)?

1. Define a `ClusterServingRuntime` whose vLLM runner serves under its own name:

```yaml
apiVersion: ome.io/v1beta1
kind: ClusterServingRuntime
metadata:
  name: vllm-demo
spec:
  protocolVersions: [openAI]
  engineConfig:
    runner:
      name: ome-container
      image: vllm/vllm-openai:v0.8.5.post1
      command:
        - python3
        - -m
        - vllm.entrypoints.openai.api_server
        - --port
        - "8080"
        - --model
        - $(MODEL_PATH)
        - --served-model-name
        - qwen2-5-0-5b-instruct      # anything other than "vllm-model"
```

2. Deploy an `InferenceService` using it, and wait for it to become Ready.

3. Run a `BenchmarkJob` against that InferenceService:

```yaml
apiVersion: ome.io/v1beta1
kind: BenchmarkJob
metadata:
  name: demo-bench
spec:
  endpoint:
    inferenceService:
      name: <isvc>
      namespace: <ns>
  task: text-to-text
  trafficScenarios: ["D(100,100)"]
  numConcurrency: [1]
  maxTimePerIteration: 1
  maxRequestsPerIteration: 10
  outputLocation:
    storageUri: "pvc://<pvc>/benchmarks"
```

4. Observe the generated Job's args contain `--api-model-name vllm-model`, the BenchmarkJob reaches `Completed`, and the per-run JSON in the output shows every request errored.

That vLLM rejects an unserved name is directly observable. Against a server started with
`--served-model-name qwen2-5-0-5b-instruct vllm-model` (the alias added deliberately, as a
workaround for exactly this bug), posting `/v1/chat/completions` with three different `model`
values:

```
qwen2-5-0-5b-instruct -> 200 OK
vllm-model            -> 200 OK
not-a-real-name       -> 404 {"object":"error","message":"The model `not-a-real-name` does not exist.","type":"NotFoundError","code":404}
```

Remove the `vllm-model` alias — i.e. configure the runtime the way anyone would who has not hit
this bug — and the second line becomes the third.

## Anything else we need to know?

One constraint that is easy to miss when fixing this, and that I got wrong on a first attempt (#779, withdrawn):

**Reading `--served-model-name` from `inferenceService.Spec.Engine.Runner` alone is not sufficient.** In the common case the flag lives in the *ServingRuntime* template, and `spec.engine` on the InferenceService carries only `minReplicas`/`maxReplicas` with no runner override. An InferenceService-only lookup finds nothing there and silently falls through to whatever fallback is chosen. Whatever source is used has to reflect the runtime template and any InferenceService-level override together, i.e. the effective engine arguments rather than either layer on its own.

For the case where a runtime omits `--served-model-name` entirely, vLLM's own default is worth matching rather than inventing a convention — `vllm/entrypoints/openai/api_server.py`:

```python
if args.served_model_name is not None:
    served_model_names = args.served_model_name
else:
    served_model_names = [args.model]
```

so the model path is the name the server will report. SGLang defaults the same way from `--model-path`.

Happy to send a PR if the direction is agreed.

## Environment

- OME version: `main` @ 23c97f19
- Kubernetes version: v1.29.15
- Cloud provider or hardware configuration: bare-metal GPU node
- OS: Linux
- Runtime and version: vLLM 0.8.5.post1
- Model being served: Qwen2.5-0.5B-Instruct
- Install method: Helm

weetime:
## What this PR does

Replaces the hardcoded `"vllm-model"` that `BuildInferenceServiceArgs` passed as
`--api-model-name` with the name the engine actually serves the model under, read from the
running engine pods.

Resolution order:

1. `--served-model-name` from the engine pods' container argv, accepting both `--flag value`
   and `--flag=value`. vLLM accepts several names for one model and reports the first back to
   clients, so the first is the one used.
2. Otherwise `--model` (vLLM) or `--model-path` (SGLang), expanded against the container's own
   environment — OME runtime templates point these at `$(MODEL_PATH)`, so the raw argument is a
   placeholder rather than the value the engine sees. This mirrors what both engines do
   themselves when `--served-model-name` is absent.
3. Otherwise an error naming the InferenceService and the flags that were looked for, instead of
   guessing.

The value is read from the pods rather than from the ServingRuntime or the InferenceService
because `--served-model-name` can be set on either one — runtimes normally carry it, while an
InferenceService may override the runner — and only the pod reflects the two merged. Reading
either spec on its own silently misses the other. The benchmark controller already holds `list`
permission on pods, so this needs no RBAC change.

## Why we need it

`--api-model-name` becomes the `model` field of every request genai-bench sends. vLLM rejects a
name it does not serve, so a runtime whose `--served-model-name` is anything other than the
literal `vllm-model` fails on every single request.

The failure is silent, which is what makes it costly: the pod exits 0, the BenchmarkJob reports
`Completed`, and a full set of result files and plots is written. Only `num_error_requests`
inside the per-run JSON reveals that nothing succeeded.

This does not affect the bundled runtimes — all 17 under `config/runtimes/vllm/` serve as
`vllm-model`, and while the 189 under `config/runtimes/srt/` do not, SGLang echoes
`request.model` back without validating it. What breaks is a user-defined vLLM runtime, which is
the normal case once you bring your own.

Fixes #781

## How to test

Unit tests covering the resolution order and its edge cases:

```
$ go test -count=1 ./pkg/controller/v1beta1/benchmark/...
ok  	sigs.k8s.io/ome/pkg/controller/v1beta1/benchmark	1.513s
ok  	sigs.k8s.io/ome/pkg/controller/v1beta1/benchmark/reconcilers/job	2.863s
ok  	sigs.k8s.io/ome/pkg/controller/v1beta1/benchmark/utils	2.278s

$ go test ./pkg/controller/v1beta1/benchmark/utils/... -run TestResolveServedModelName -v
--- PASS: TestResolveServedModelName
    first_of_several_served_model_names_wins
    equals_form_is_accepted
    falls_back_to_--model_with_$(VAR)_expanded
    falls_back_to_SGLang_--model-path
    a_flag_directly_after_--served-model-name_is_not_its_value
    pods_of_other_inference_services_are_ignored
    no_engine_pods_is_an_error
    engine_pod_naming_no_model_is_an_error
```

End to end: deploy an InferenceService on a vLLM runtime whose `--served-model-name` is not
`vllm-model`, run a BenchmarkJob against it, and check the generated Job's args. Before this
change they carry `--api-model-name vllm-model` and every request 404s while the job still
reports `Completed`; after it they carry the served name and the requests succeed.

That vLLM rejects an unserved name is directly observable — against a server started with
`--served-model-name qwen2-5-0-5b-instruct vllm-model`:

```
qwen2-5-0-5b-instruct -> 200 OK
vllm-model            -> 200 OK
not-a-real-name       -> 404 {"object":"error","message":"The model `not-a-real-name` does not exist.","type":"NotFoundError","code":404}
```

## Checklist

- [ ] Docs updated (if applicable) — N/A, no user-facing API or behaviour to document beyond the fix itself
- [ ] `make test` — did not run the full target: `pkg/xet` needs its Rust bindings built and the
      envtest suites need `/usr/local/kubebuilder/bin/etcd`, neither available in my environment.
      Both fail identically on a clean `main`, so this change introduces no new failures. I ran
      `./pkg/controller/...` on a clean `main` and on this branch and compared: 37 `ok` on both,
      with the same 5 pre-existing failures on each.

## Summary by CodeRabbit

* **Bug Fixes**
  * Benchmark requests now use the model’s configured served name instead of a fixed default.
  * Supports served names specified through command-line options, model paths, aliases, and environment variables.
  * Improved compatibility with different engine configurations, including SGLang.
  * Added clearer error handling when model information cannot be determined or engine details are unavailable.

### DIFF
--- a/pkg/controller/v1beta1/benchmark/utils/utils.go
+++ b/pkg/controller/v1beta1/benchmark/utils/utils.go
@@ -3,12 +3,14 @@ package benchmarkutils
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/ome/pkg/apis/ome/v1beta1"
+	"sigs.k8s.io/ome/pkg/constants"
 	isvcutils "sigs.k8s.io/ome/pkg/controller/v1beta1/inferenceservice/utils"
 	"sigs.k8s.io/ome/pkg/utils/storage"
 )
@@ -66,9 +68,14 @@ func BuildInferenceServiceArgs(ctx context.Context, c client.Client, endpointSpe
 			return nil, fmt.Errorf("BaseModel %s has missing Storage or Path information", baseModelName)
 		}
 
+		modelName, err := resolveServedModelName(ctx, c, inferenceService)
+		if err != nil {
+			return nil, err
+		}
+
 		args := map[string]string{
 			"--api-key":         "sample-key", // TODO: Use actual service account key later
-			"--api-model-name":  "vllm-model",
+			"--api-model-name":  modelName,
 			"--model-tokenizer": *baseModel.Storage.Path,
 		}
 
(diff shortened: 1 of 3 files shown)


pallasathena92 (approved):

pkg/controller/v1beta1/benchmark/utils/utils.go
```diff
@@ -106,6 +113,93 @@ func BuildInferenceServiceArgs(ctx context.Context, c client.Client, endpointSpe
...
+			isvc.Namespace, isvc.Name, err)
+	}
+
+	var modelLocation string
+	for _, pod := range pods.Items {
+		for _, container := range pod.Spec.Containers {
+			argv := append(append([]string{}, container.Command...), container.Args...)
+
+			if name := flagValue(argv, servedModelNameFlag); name != "" {
+				return name, nil
```
pallasathena92, line 150: nit. 
`--served-model-name` isn't expanded here, but the `--model` fallback below is.

Three bundled runtimes pass it through an env var — `config/runtimes/vllm/llama-3-1-8b-instruct-rt.yaml:58`, `llama-3-3-70b-instruct-rt.yaml:61`, `llama-3-3-70b-instruct-fp8-dynamic-rt.yaml:59`:

```yaml
- --served-model-name
- $(SERVED_MODEL_NAME)
env:
  - name: SERVED_MODEL_NAME
    value: "vllm-model"
```

`flagValue` returns the literal `$(SERVED_MODEL_NAME)`, so genai-bench sends `{"model": "$(SERVED_MODEL_NAME)"}` and vLLM 404s every request — the same silent failure this PR fixes. These three work today only because the hardcoded `vllm-model` is the right answer for them.

```go
if name := flagValue(argv, servedModelNameFlag); name != "" {
    return expandContainerEnv(name, container.Env), nil
}
```

Worth a test too — the existing `$(VAR)` case only covers the `--model` branch.

PR #791, opened 2026-08-26
Optimize multipart model downloads with direct writes

fredericsun:
## What this PR does

- Replaces multipart per-part temporary files and serial assembly with direct `WriteAt` calls to non-overlapping offsets in one preallocated object-level temporary file.
- Uses a fixed 1 MiB coalescing buffer to avoid small disk writes.
- Adds a Pod-wide write concurrency limit to control NVMe pressure.
- Publishes the file through atomic rename only after all parts complete successfully.

## Why we need it

The previous multipart path wrote model data twice: first to `/tmp` part files, then into the assembled model file. It also caused part workers to wait behind a single assembler.

Testing showed that direct writes:

- Reduced physical writes from approximately 2× to 1× model size.
- Removed multipart `/tmp` usage and assembler backpressure.
- Reduced single-model Bulk download time by 21.8–30.2%.
- Reduced two-model Bulk makespan by 43.2%.

## How to test

```bash
go test ./pkg/ociobjectstore ./pkg/modelagent ./cmd/model-agent
go test -race ./pkg/ociobjectstore
helm template ome-resources charts/ome-resources --set modelAgent.enabled=true
```

## Checklist

- [x] Tests added/updated
- [x] Configuration and inline documentation updated

## Summary by CodeRabbit

* **New Features**
  * Added configurable limits for concurrent model-file writes and downloads, defaulting to 8.
  * Added Helm configuration with validation requiring a positive concurrency value.
* **Performance**
  * Improved parallel model downloads by writing directly to preallocated files.
  * Reduced temporary-file usage and improved retry, cancellation, and storage-allocation efficiency.
* **Bug Fixes**
  * Improved handling of short responses, partial retries, and file-part boundaries.
* **Tests**
  * Added coverage for concurrency limits, retries, boundary handling, and invalid settings.


pkg/ociobjectstore/write_limiter.go
```diff
@@ -0,0 +1,33 @@
+package ociobjectstore
+
+// WriteLimiter bounds concurrent model-file writes across all objects and
+// model download tasks in one model-agent process. A nil limiter is unlimited.
+type WriteLimiter struct {
+	permits chan struct{}
+}
+
+func NewWriteLimiter(concurrency int) *WriteLimiter {
+	if concurrency < 1 {
```
truddy0, line 13: do we want to guard this? if the concurrency is set as 0 accidentally, the limiter will be unlimited

fredericsun: Good catch. 0 was initially used to preserve the previous unbounded-write behavior when the option was not configured, but that weakens the safety boundary of the limiter. I changed the CLI and Helm defaults to 8 and added validation in both the model-agent and Helm chart. 0 or negative values now fail fast instead of enabling unlimited writes.

pkg/ociobjectstore/os_parallel_download.go
```diff
@@ -303,37 +299,23 @@ func (cds *OCIOSDataStore) downloadFilePart(ctx context.Context, prepareDownload
...
-
-				// Stream data directly to temp file using pooled buffer
-				bufp := BufferPool.Get().(*[]byte)
-				written, streamErr := io.CopyBuffer(tempFile, resp.Content, *bufp)
-				BufferPool.Put(bufp)
-
+				// Stream this range directly into its disjoint offset in the shared
+				// object-level temporary file. OffsetWriter uses WriteAt and does not
+				// mutate the shared file cursor.
+				written, streamErr := writePartAtWithLimiter(
```
truddy0, line 317: The new LimitedReader silently truncates oversized responses without checking whether GetObject honored the requested range. If the server returns 200 OK with the whole object, each worker writes the object’s beginning at its assigned offset and reports success.

fredericsun: Thanks @truddy0. This is a nice precaution. I've updated the commit to add an extra validation for the range response.

PR #838, opened 2026-09-01
Prioritize serving-demand model downloads

gsgvijay:
## What this PR does

- Prioritizes nodel-local model downloads that are required by active InferenceServices, without interrupting downloads that are in progress.
- Adds endpoint-demand awareness to the model-agent and a bounded, multi-priority task scheduler.
- Allows a model required for serving to advance ahead of background cache warming and startup revalidation work.

## Why we need it

- The model-agent can receive a large amount of background model work when a node cold-starts or when the model catalog changes
- Model needed by an InferenceService can therefore wait behind unrelated downloads even though endpoint readiness requires it
- Existing task queues have no configurable upper bound and a large burst of model events can retain an unbounded amount of queued work per node

### Prioritize demanded models

Download and DownloadOverride tasks carry a ServingDemand marker when the referenced model has current endpoint demand

Demanded tas:
- Replaces queued background work for the same model UID
- Runs on high-priority worker queue
- Can displace queued normal revalidation or background download task when the queue is full
- Does not displace delete work or other high-priority work
- Does not cancel or restart a download that is in-progress

Existing delete and same-path reuse priority remain unchanged.

### Bound queued work

Task scheduler now has a configurable max number of distinct queued tasks (defaults to 4096). Normal downloads and startup revalidation are rejected when the queue is at capacity.

A serving-demand task can make room by removing lower-priority queued work.

### Configuration and deployment wiring

model-agent binary adds:
- `--demand-priority-enabled` (default: false)
- `--task-scheduler-capacity` (default: 4096)

Helm and raw manifest paths expose the scheduling settings and grand model agent read-only access to InferenceServices. The helm chart continues to leave the entire model-agent DaemonSet disabled by default. When a deployment enables the DaemonSet, it can independently choose whether endpoint-demand prioritzation is enabled.

## Validation

```
go test ./pkg/modelagent ./cmd/model-agent
```

Result:
```
ok sigs.k8s.io/ome/pkg/modelagent
ok sigs.k8s.io/ome/cmd/model-agent
```

## Summary by CodeRabbit

- **New Features**
  - Added configurable model download priorities: Background, Standard, and High.
  - Added scheduling status for serving demand and reference counts.
  - Serving demand can automatically prioritize required model downloads.
  - Added configurable worker counts, queue capacity, and same-path reuse timeout.
  - High-priority downloads can displace lower-priority queued work when capacity is reached.

- **Changes**
  - Model-agent pods no longer default to system-critical priority.
  - Replaced the same-path wait-timeout setting with the reuse-wait-timeout setting; the previous option remains available as a deprecated alias.


charts/ome-resources/values.yaml
```diff
@@ -387,6 +387,27 @@ ome:
       affinity: {}
       priorityClassName: ""
 modelAgent:
+  # Deploy the model-agent DaemonSet (node-local model download/cache).
+  # Required for BaseModel/ClusterBaseModel-backed model resolution.
+  # Installs whose runtimes fetch models themselves can leave this off.
+  enabled: false
+  hostPath: /mnt/data/models
+  priorityClassName: system-node-critical
```
slin1237: don't make ome agent as system critical. allow user to set this

charts/ome-resources/values.yaml
```diff
@@ -387,6 +387,27 @@ ome:
...
+  serviceAccountName: ome-model-agent
+  numDownloadWorkers: 2
+  numHighPriorityWorkers: 1
+  taskSchedulerCapacity: 4096
+  samePathWaitTimeout: 30m
```
slin1237: What does same path wait time out mean? It's a bit confusing

charts/ome-resources/values.yaml
```diff
@@ -387,6 +387,27 @@ ome:
...
+  demandPriorityEnabled: true
```
slin1237: Likewise, it's hard to understand the meaning of this flag. 
This is essentially priority scheduler right?

config/model-agent/clusterrole.yaml
```diff
@@ -20,3 +20,6 @@ rules:
   - apiGroups: [ "ome.io" ]
     resources: [ "clusterbasemodels" ]
     verbs: [ "get", "list", "watch", "patch", "update" ]
+  - apiGroups: [ "ome.io" ]
+    resources: [ "inferenceservices" ]
```
slin1237: I'm a bit concerned about this
this means daemonSet will watch over all isvc
this is super heavy against k8s itself and gets worse as cluster size grows

slin1237 (changes requested):
I don't think this is the right path to enable priority scheduling for model agent
Model agent should only be concerned about model weight downloads and watch over model resource
If we wanna enable priority scheduling, perhaps allow model controller to dynamically set priority or allow user to set priority as part of model.storage spec

charts/ome-resources/values.yaml
gsgvijay: It's for when another model task is already populating the same local artifact path. The second one waits so it can reuse the completed files. If the first task doesn't become Ready within the configured duraiton, it resumes the normal download flow

charts/ome-resources/values.yaml
gsgvijay: Yes, it controls only model-download queue ordering inside model-agent. When enabled, a model referenced by InferenceService receives serving-demand priority and is placed in teh high-priority queue.

config/model-agent/clusterrole.yaml
gsgvijay: You're correct, this could have a cluster-wide InferenceService watch. Would this approach work better:
- Have 1 leader-elected OME controller watch InferenceServices cluster-wide
- Maintain demand reference counts by UID
- Project a controller-owned `serving-demand` marker into the referenced Cluster/BaseModel
- Write on `0 to 1` or `1 to 0` demand transitions
- Let model-agent consume this through its already existing model informers
- Remove InferenceService RBAC and informer from each model-agent pod

gsgvijay: Fixed it. Model-agent no longer watches `InferenceService` resources and no longer needs teh RBAC. A leader-elected model-controller now owns the cluster-wide watch and uses a field index to find endponts referencing a BaseModel/ClusterBaseModel

charts/ome-resources/values.yaml
gsgvijay: Removed `demandPriorityEnabled` from model-agent and added a controller-owned `servingDemandPriorityEnabled` for readability.

charts/ome-resources/values.yaml
gsgvijay: Renamed to `samePathReuseWaitTimeout` for readability. It is the max time a model task waits when another task is already populating the same local artifact path. If the other task doesn't establish a reusable `Ready` artifact before timeout, the waiting one will resume the normal flow.

charts/ome-resources/values.yaml
gsgvijay: The model-agent no longer receives a system-critical PriorityClass by default. The `priorityClassName` defaults to empty and the DaemonSet renders it only when an operator explicitly supplies it.

pkg/modelagent/gopher.go
```diff
@@ -203,7 +214,20 @@ func (s *Gopher) enqueueTask(task *GopherTask) {
...
+		return
+	}
+	pending := task
+	if result.accepted {
+		pending = result.displaced
+		s.logger.Debugf("Deferring displaced model-agent task until scheduler capacity is available: %s", getModelInfoForLogging(pending))
+	} else {
+		s.logger.Debugf("Waiting for scheduler capacity for model-agent task: %s", getModelInfoForLogging(pending))
+	}
+	if !s.taskQueue.enqueueWhenAvailable(pending) {
```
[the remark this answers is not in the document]
gsgvijay: Replaced blocking deferred admission to scheduler owned pending queues and made dispatcher submission return immediately after recording hte task. Pending work is also now coalesced by model id and UID.
`endqueuWhenAvailable()` is removed from dispatcher path.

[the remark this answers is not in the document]
gsgvijay: Pushed the new changes

pkg/modelagent/gopher_task_queue.go
```diff
@@ -8,44 +8,170 @@ import (
...
-	} else {
-		q.normalDownload = append(q.normalDownload, task)
 	}
-	q.cond.Broadcast()
+	// Deferred work is coalesced by object UID for every task class. This keeps
+	// scheduler-owned overflow proportional to model cardinality rather than
+	// informer event volume while retaining same-name recreations with new UIDs.
+	q.pendingHigh = removeSupersededTasks(q.pendingHigh, task)
+	q.pendingDownload = removeSupersededTasks(q.pendingDownload, task)
+	q.pendingRevalidation = removeSupersededTasks(q.pendingRevalidation, task)
```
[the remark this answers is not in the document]
gsgvijay: Added queue coalescing to preserve stronger pending download intent against stale retries

pkg/openapi/swagger.json
```diff
@@ -7427,6 +7446,10 @@
           "description": "DownloadPolicy describes the policy of downloading model artifacts Supported policies: - AlwaysDownload: always download a copy of model artifact in destination path - ReuseIfExists: if the identical model artifact has been downloaded in the node, such artifact will be reused",
           "type": "string"
         },
+        "downloadPriority": {
+          "description": "DownloadPriority is the persistent user-selected priority for materializing this model on eligible nodes. Serving demand observed by the model controller can raise, but never lower, this priority.",
+          "type": "string"
+        },
```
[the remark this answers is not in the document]
gsgvijay: API scehma and generated OpenAPI files enforce the priorityenum

slin1237 (approved):

op109lvb (changes requested):
We should keep actual downloads out of the high-priority worker pool, which is reserved for delete and reuse/copy operations.

`runHighPriorityWorker()` calls `processTaskWithOptions(task, false)`, but the Hugging Face download path doesn’t honor that restriction. A high-priority HF download can therefore occupy the worker and block unrelated deletes. Putting deletes at the front of the queue doesn’t help when all workers are already downloading.

Can we prioritize serving-demand downloads within the download worker pool while preserving dedicated capacity for delete/reuse work? Endpoint download priority shouldn’t come at the cost of blocking cleanup.

gsgvijay:
> We should keep actual downloads out of the high-priority worker pool, which is reserved for delete and reuse/copy operations.
> 
> `runHighPriorityWorker()` calls `processTaskWithOptions(task, false)`, but the Hugging Face download path doesn’t honor that restriction. A high-priority HF download can therefore occupy the worker and block unrelated deletes. Putting deletes at the front of the queue doesn’t help when all workers are already downloading.
> 
> Can we prioritize serving-demand downloads within the download worker pool while preserving dedicated capacity for delete/reuse work? Endpoint download priority shouldn’t come at the cost of blocking cleanup.

Fixed it. Download priority now orders High > Standard > Background within the download worker pool; delete/resuse workers retain dedicated capacity.

HF and OCI paths (including overrides) enforce no-remote-download boundary on cleanup workers. Reuse misses transfer to download workers without losing task identity or priority, and displaced remain queued.

